PR #1, opened 2017-08-23
EN server support

ryuga93:
Added EN server support


29988122:
Thank you for contributing! I thank you on behave(?) of all the EN server players.

By the way, beware *not* to post this on reddit / popular sites. Only smaller ones or private msg maybe.

One reason is that I asked mod and he said although CCC event is fucking farming intensive, he still thought this is not suitable for the sub. 

Another reason is that Delightworks can simply blocked the daemon process used here, and I'll have to ask the author of Ankulua framework to deal with it - and it might be impossible to do so just like ANRC's development currently on hold.

Ahh, but I guess EN server is free from this root blocking crap? : p

ryuga93:
I agree. Let's keep this low profile. I think they are heavily against automation, cuz emu has the highest chance to automate the game, hence they block them. EN server is the same in blocking root/jailbreak. Can't run the game even just with USB debug mode on...

I used to play on JP server until last year when they decided to block jailbreak devices. My customization on Ipad has more values than the game so I just quit :p

29988122:
@ryuga93 

Hi, I noticed the option setImmersiveMode(true) has been set by you in the English version.

How's the virtual button row affecting FGO? I'd like know the behavior, and if setImmersiveMode(true) affects the script performance - since this makes every compare action done twice.

ryuga93:
I haven't test it without setImmersiveMode. When I first used your script, all of the coordinates are wrong, but that could be because I changed the setScriptDimension to my resolution instead of 2560.

However, my phone (Xperia Z2) has virtual button row and the game will fill up the whole screen. Using the Ankulua calculator script, the yellow touchable region will only cover 1776 pixels instead of the full 1920 pixel without setImmersiveMode. So I just remain it there even after changing back setScriptDimension to 2560.

A better solution is to create a menu for user to select, but I think since it is open source, anyone can change it once in the code and run the script directly without going through the menu every time.

29988122:
Thank you for the info, it helps when I encountered virtual button issue in the future!

By the way you don't really have to change scriptDimension - it's gonna auto adjust to your screen resolution, that's only useful for script writing - you don't want to calculate for coords everytime you grab a new pic for recognition.

ryuga93:
Yup, thanks for that. I noticed the fact after looking at the Chinese doc. English doc did not have much detailed documentation.

PR #15, opened 2018-01-02
Better Chaining Module

ryuga93:


ryuga93:
Let me know if this chaining is better

29988122:
That is....very elegant. 
Much more better that what I have done.

As a newbie, I still have some very bad habits like..global variables(bcs Lua..also why the array started from 1 not 0), bad naming, exceptions, error handling, etc.

Reading your codes makes me learned a lot. Especially when I was learning python, I have no idea what an iterator could do. Now I know. 

Although there's still some naming inconsistency in MY code, like Card1Region should be Card1AffinRegion..whatever, might as well leave it alone for now), I should approve the pull request for obvious reasons.

Thank you Mr. Ryuga93!

=-=-=
I think it's a very good foundation for us to add other functions onto it. Like servant / master skill using that you've mentioned, could be useful for someone who wants to craft their unique efficient event 周回 team.
(What do you call 周回 in English? Grinding/farming?) 

Few quick questions:
1. The sequence of storage array is the priority of card choosing, right? 
2. Does affin stands for affinity? Like, 相性？(Saber->Lancer->Archer->Saber)
3. What is the reason behind choosing resist Buster cards? Does it have greater dmg than normal Q, A cards? Never dig into those dmg calculations, so I have no idea.
4. In rare conditions when user got the wrong team, therefore we have all resist A Q cards on board, it will stuck, right? (user will have to stop the script to switch team anyway, the behavior should be OK.)

5. Took me a while to read the double iteration part, still have no idea why you were using ipairs() and pairs() respectively. Any keywords for me to google? Thanks!
`` 
	for i, storageNum in ipairs(storage) do
		for j, nCard in pairs(storageNum) do
			click(CardClickArray[nCard])
			cardCount = cardCount + 1
			if cardCount == 3 then
				break
			end
		end
	end	
``

ryuga93:
Hi, I am now 90% finish on the autoskill, once it is ready I will push up along with some small bug fix for this module.

1)Yes, basically we can alter the storing sequence to give different priority, eg: to focus on buster/arts/quick, that would give a lot of customization for users based on their team, but I will do it when someone request it, or I feel I need to use it :p

2)Yup, but basically it just identify the word "weak"

3)The damage data is [here](http://fategrandorder.wikia.com/wiki/Command_Cards)
The resist card has 0.5 multiplier, and Buster has 150% damage, so a resist buster has 75% damage which is slightly below quick(80%) and arts(100%). 

But the real deal is its effect as first card towards the following two cards, giving them +50% increase in damage, which makes up the lost. Besides, getting a buster chain will add 20% to each card, making it 95% if placed first, but  normally it will fall on the 3rd card based on the code, so it is 105% with 20% increase = 125%. The 3rd arts card has 140% increase, but don't forget the buster chain effect on the other two cards which has 20% increase each, that's 40%-(140-125)%=25% more damage compared to 3rd arts.

4)Yep, that's the bug fix that I will push along with Autoskill

5)If you look up the table/array in [Lua documentation](http://lua-users.org/wiki/TablesTutorial) you can see that Lua array is quite special that it can act as object too. But think it as a table, where the table has the following:
Index | Value
1        |  14
3        |  10
a        |  23
2        |  22
4        |  43

Now let's use ipairs
`Result:
Index: 1, 2, 3, 4 
Value: 14, 22, 10, 43`

While for pairs
`Result:
Index: 1, 3, a, 2, 4
Value: 14, 10, 23, 22, 43`

So, ipairs is a more strict index ordering that includes numbers only, while pairs just follow along the index and read all the value. As a rule of thumb, use ipairs for arrays and pairs for tables

And the special stuff about Lua array is the array can store arrays/tables in it. Meaning:
Index | Value
1        |  14
3        |  {1, 2, 3}
a        |  23
2        |  22
4        |  43

And back to the code, where I explicitly define the index: 
Index        |       Value
     1          | WBLocation
     2          | BLocation
     3          | WALocation
     4          | WQLocation
     5          | RBLocation
     6          | ALocation
     7          | QLocation
where each value is actually tables, and since I would not want to mess the storage sequence, ipair will do the job. 

Now let's look at WBLocation:
Index        |       Value
     1?        | 5
     2?        | 3
The index is unsure since I used table.insert, but generally it will be ordered numerically from 1, I used pairs just to prevent missing value or wrong sequence in case the index is not ordered numerically.

29988122:
Got it, thx bro for teaching me!

ryuga93:
Lua array accept negative index too, makes life easier for the autoskill function

PR #19, opened 2018-01-04
Support Autoskill & Bug Fix

ryuga93:


29988122:
Reading the code, guess I need to get basic understanding about string processing(regular expression and your way of handling exceptions)

Thank you again!

ryuga93:
Lua uses Pattern, a bit different from Regex.

I used the knowledge from Java and ended up wondering why it didn't work as I wanted initially...

Issue #29, opened 2018-01-21
Pick waver and buff all allies


MerryNight:
You can modify the script to autopick a support waver and enable to use the skill buff all allies from chaldea code? Thanks for read. 10/10 script

MerryNight:
Now I realized how to use the master skills. However, I still want to autopick waver.

29988122:
Sorry, can't really understand what you mean of "autopick waver"

Can you try to explain it?

aqer:
He wants to autopick servant "Waver" when choosing support servant.

29988122:
Oh....got it, language barrier.
Our cute little Lord El-Melloi II in Zhuge Liang, right?

It definitely can be done, but since I need to figure out how to do it properly, could take a while. 
Lots of exception handling and screenshot has to be done. 
So soon™

In the meantime, you can try contribute some $$$ to Delightworks and Typemoon, and hope to get either Merlin or Waver for yourself : -

Jared-Lee:
如同類似這篇的需求，希望有一個功能可以自動選取加絆禮裝(英靈肖像/ 學妹 午餐 カルデア・ランチタイム)來對應在FREE本減半的時候可以刷低AP的關卡來刷絆。
無論如何，感謝了。

29988122:
I guess the flow can be executed like:
1. user screenshot and crop servant/Reisou that you wanna choose
2. put it in folder
3. modify config to enable this function
4. script will search the first page of support list for it; if not existed, execute as normal.

29988122:
@MerryNight @Jared-Lee 

Now it's supported in version 2.0.0. 
Check readme for further info!

PR #42, opened 2018-06-01
Added "Support_SelectFirstVisible" option

potchy:
When 1, automatically select the first visible support (it's the current behavior).
When 0, pause at the support selection screen, vibrate for 3 seconds and wait for user input.

During events, there are, usually, specific Servants and Craft Essences that increase the drop chance of event currency. When that is the case, like in the "GO WEST!" event that is currently ongoing in NA, I prefer to select the support servant myself, so I'm always running the most optimal farming setup.


29988122:
While implementing a new function(in this case, a *proper support servant/礼装 feature*, I need to consider:

1. The "automation level" of the script, i.e., for how long can it execute without human intervention.

2. Exception handling, e.g., in this branch https://github.com/29988122/Fate-Grand-Order_Lua/tree/roundtest I'm testing a new way to detect current stage among different regions, which is not very reliable and often misjudge for now. That's why I haven't merge it into master yet. 

A lot of the cases were rare behavior like servant kizuna lvl up, FGO server pushed an update or somehow it bugged up and shows your daily login bonus window...Gotta handle them as much as possible. 
Hell, sometimes even when you hit refresh support list like a dozen times, it could still be empty especially in the first/last day of the event if you choose to filter only event related servants.

3. Compatibility. Bilibili(CN) and JP server often were on different versions, hence different font and UI flow / behavior. 

4.Usability. How can the script simply its configuration file so regular user can use it without any issue.

While I'd like to implement a function to let user predetermine a priority list for their preferred support servant/礼装, it's a little bit troublesome considering all the four points mentioned above.
Yeah the code can be done quickly by a veteran, or by scanning memory(the PC way, not practical on Android), but I'm just a noob which makes it challenging for me. 

Discussion / suggestions are welcomed, thank you potchy.

potchy:
Hmm, I admit that waiting for user input is not very practical. I'm not expecting you to merge it right away, really, but I believe this is a good alternative for people who want to select the most optimal Servant/Craft Essence for farming, while issue #29 isn't solved. 

I'm also just starting with Lua scripting, but I have many ideas for this, so I'll be submitting a lot of pull requests for review. Be prepared. :D

29988122:
Just drafted a new release for autoskill. Looking forward to your help : D

TryBane:
You currently have a method of letting the user prioritize an enemy unit over another during events if I recall correctly. I'm not a Lua person myself, but if I understood your code correctly, you take an image provided by the player and use that to select what servant to prioritize during battle.

Can't you theoretically request the player to provide an image of a CE/Servant (or both) they want in their support slot?

You can provide 4 options:

1. First selectable
2. Servant Priority
3. CE Priority
4 Servant/CE Priority

Later you can expand this so that an array of CEs and/or Servants are prioritized. I doubt you want to go so far to order the priority though.

But now you have the user name the file "Priority_Support.png" If you provide them the option of an array, you can have them name them accordingly eg "Priority_Support_1.png" And have them follow certain crops as you request of them for the Servant Priority for battle currently. These may include different crops for Servant vs CE vs Servant/CE. This should also be easily explained as well in your configuration file and your read.me file.

Now for the more difficult part for you, you'd need to read the servants that are available to determine if they match any of the priorities. In the case of not being able to find one, then you need to program it to refresh the servant list and search again. Some cases to be aware of before you start is to not continue scanning for servants beyond the actual friends since if the user needs to use a friends' NP, the script won't work for those units.

You may want to implement a timeout as well, where if the priority support doesn't show up after x amount of refreshes (which you may provide as an adjustable option for your users) then you revert to the first option for that run.

I don't believe I missed anything but I could very well be wrong.

ryuga93:
I wrote an initial draft of the functions you have stated a few months ago, but to run it as accurate as the user want, the matching must be 100%. The few factors I put off the development are:
1) The final art accession setting/ the intermediate assension art. Sometimes the servant's art will be quite similar at different accession stages, making the recognition to fail to select the actual one, unless there is a method to do 100% match. 
2) The limit break CE, which can only be differentiated by the small star symbol. Same as 1), unless there is a 100% match method.
3) The auto scrolling and checking loop, as sometimes the support you want does not appear in the top list. This is hardest part, as the script needs to know how much pixels to scroll.

Appart from those, my initial draft worked, where it can refresh the list and select the servant/CE I want, but not 100% matching the criteria, such as limit break CE or servants with higher state/skill level. Also I need to sort the list in a certain order to make the support I want appear in the top list. Though I can make a strict search by using the username image, still there might be other user with the exactly same support I want in the list, so the script will not be efficient time wise. TBH I feel that this requires more work than the card priority and autoskill functions as a lot of documentation reading needs to be done XD.

TryBane:
I've been reading the documentation as well, and here's what I've got.

You probably know this, but you can use the find() function to find an image and set it to a Match object, which can then be used for the click() function. And instead of searching for the art of the servant, we can search for their name. Allow the user to use the name as the requirement and there shouldn't be much of an issue in terms of art.

As for the CE vs MLB CE, I imagine that you can include in the document that the image provided must contain the little star symbol on the right if they specifically want the MLB version. However, sometimes non-MLB versions are acceptable for the user, and in those cases a partial image that doesn't incorporate the region where the star is can be used for the scanning.

I was about to create my own draft as well to see how well this works. I was going to use the art as you mentioned until you told me the various problems there are with doing so. And so I realized that we can use the name instead. This will also potentially eliminate the possibility that you choose a non-friend support since the contrast change may keep those supports from being selected.

And as for the scrolling issue, using the region in which the find() function returned should help with needing to be exact. Just provide a scroll of half the screen every interval of searches, update if none is found, and repeat till it is found. I don't believe it should scroll through the whole friends list though, since doing so will firstly require more time (even more time wasted if you don't have many friends meeting filter criteria during events)  and since most people search based on health, atk, or lvl in descending order, going to the bottom of the list will not be optimal in most scenarios.

If any of this gives you ideas of how to implement it, that'd be great. My scenario where I need this is to get a friend with MLB Mona Lisa CE during QP farming, which is why I would really like to get it set up :D

One last thing however, I see that there may be a bigger pain with trying to implement Servant AND CE filtering, so we may be able to search for the name of the servant first, then check for a CE that matches as well relative to the location of the name. This may be difficult to set up properly though. I imagine it would include first finding if they both exist on screen, then taking the region the name was found, expanding the region size left and down a bit, and checking that region to see if the CE image exists. And when that is the case, you've found your match.

Let me know if you find this helpful. I will go ahead and see if I can get it implemented myself and will post a new pull request if I find it feasible.

29988122:
So 礼装 is called CE = Craft Essence in English version....interesting. I find it hard to translate but craft essence is good enough.

I only have one added concern:
"Allow the user to use the name as the requirement and there shouldn't be much of an issue in terms of art."

It'd be a pain in the ass if other localized version(tw, cn to be exact) do something weird in this column -  typography, position, or other weird things, ever after toned down the tolerance of find() or match() it could still fail.

I know you're discussing user taken photos for this feature, just a side trivial concern.

For example, I still don't know why this happened, and IRRC I did not alter any codes related to it..
https://github.com/29988122/Fate-Grand-Order_Lua/issues/46#issuecomment-406103776
It still worked here on the JP server. 
(By the way, I broke my successive login streak for two weeks * cry * all those stones...)

Anyway I appreciated all the purposed draft, it's way above my level of coding / planning ability, so I guess I can only watch and learn from you guys.
*FYI, I'm currently reading this from the beginning to strengthen my base  - http://www.openbookproject.net/thinkcs/python/english3e/foreword.html *

29988122:
I was doing some...if that can be called...refactoring.

Hopefully the mainloop will be more clean and more easy to add functions.

potchy:
Guys, based in @TryBane's ideas, I got something that's working pretty well, but there is a problem with support refresh that may require taking screenshots of other regions besides NA, so I prefer discussing it before updating the pull request.

Scrolling is done like this:
image

If the desired servant/CE is not found, the script clicks here:
image

Then here:
image

As of now, the code is structured like this:
```
if Refill_or_Not == 1 and StoneUsed < How_Many then
    RefillStamina()
end

selectDesiredSupport() <- this is were the magic happens
wait(1.5)
```

Now, imagine you ran out of AP and you do not want the script to refill it for you.
Instead of the support selection screen, this dialog will pop up:
image

When calling `selectDesiredSupport()`, the script assumes the game is in the support selection screen, but in this case, it is actually in the SQ/apples screen. Because the "Update list" and "Yes" buttons are located more or less in the same X/Y coordinates of the SQ, the script will click them and SPEND YOUR PRECIOUS STONES!

image

What I suggest is changing the code to something like this:
```
if Refill_or_Not == 1 and StoneUsed < How_Many then
    RefillStamina()
end

if SupportScreenRegion:exists("support_screen.png") then
    SelectDesiredSupport()
    wait(1.5)
end
```

But I only have the support_screen.png for NA.
Is it too troublesome to take screenshots of other regions?
image

ryuga93:
Maybe you can try to check for the class icons instead

29988122:
I agree, and by looking at this pic you'll notice something:


Delightworks tend to change UI elements to make it *more crowded and ugly*.

That included more class icons and extra buttons added in the future(like, event bonus filter and Alter Ego/Moon Cancer/Foreigner did not even exist in this pic, it's 2016.12 version of the client.).

Therefore, I will personally choose the **arrow** of "Class change" or "Back to last page" on the top-left corner as target, but test is definitely 
test
needed to see if it's reliable enough.

potchy:
support_screen

So, I went with @29988122's idea and ended up using the image above.
I'll be testing the script in the Pryllia event coming this week in NA. If it works consistently, I'll be pulling the code this weekend. 😄

TryBane:
Can I get a copy of the script you're using for this? I'm playing on NA and would like to help confirm whether it's working if at all possible.

potchy:
@TryBane, see [my fork](https://github.com/potchy/Fate-Grand-Order_Lua/tree/potchy).

I've pushed my first take on this feature.
There are **5** new options to control support selection behavior.

---

`Support_SelectionMode` can be set to:
- **first**: the default mode everyone is used to. It will pick the first visible servant on the screen.
- **manual**: the script will stop when it reaches the support selection screen (my initial idea with this pull request).
- **preferred**: the script will pick the servant/CE specified in the `Support_PreferredImage` option.

---

`Support_PreferredImage` is where you input the image you want to use for servant/CE detection.
The image needs to be placed inside the "image_SUPPORT" folder. It can be anything as long as it is inside this area:

screenshot_20180819-122510

### Examples:

**CE only:**
angelica
miyu
tanaka

**Servant only:**
waver4

**Servant + CE:**
waver_ce

**Servant + CE + NP level:**
waver_ce_np

---

If the servant/CE is not found in the support list, the script will click refresh as many times as defined in `Support_MaxRefreshes`.

If this limit is reached, the script will use the `Support_FallbackTo` option to decide what do next. The options are the same as in `Support_SelectionMode`. For instance, if the script is setup like this:
```
Support_SelectionMode = "preferred"
Support_PreferredImage = "waver4.png"
Support_MaxRefreshes = 3
Support_FallbackTo = "first"
```
.... it will try to find Waver in your support list. If it is not found in 3 refreshes maximum, the script will scroll back to top and pick the first visible servant.

---

The last option, `Support_SwapsPerRefresh` controls the number of swaps before refreshing the screen.
I have considered to just make the script scroll until the end of the support list is reached, but we also have to remember these two things:
- The "refresh list" button cannot be spammed. The game only allows refreshes once every ~10 seconds. This option will help keeping the script busy during this time.
- The only way to differentiate **friend** servants is the gray/white NP name next to their icons. I still haven't found a way to reliably do that. We can use this option to limit the search area until then.

---

That's it, guys.
I'm all open to suggestions. 😄

29988122:
Guest I need to update readme.md XD

Thx a ton bro!

Issue #36, opened 2018-04-09
How to prioritize certain order of card selection.


MerryNight:
This. I want to prioritize arts card instead of buster cards.

29988122:
solved here: https://github.com/29988122/Fate-Grand-Order_Lua/commit/4ab656b31759f5a9a5ed73eaeccfeafcebc1c31b

Thank you potchy!

PR #44, opened 2018-06-24
Added cards priorization

potchy:
My computer died, so this feature got delayed a bit. 😄 

In response to #36, this commit introduces a new option: `Battle_CardPriority`.
The order of the cards can be setup in any way the user wants it:
- Buster, Arts, Quick (default)
- Arts, Buster, Quick
- Quick, Buster, Arts
- etc.

Effective cards are automatically placed first.
For example, considering `Battle_CardPriority` is set to "BAQ", cards will be priorized like this:
Effective Buster -> Buster -> Weak Buster -> Effective Arts -> Arts -> Weak Arts -> Effective Quick -> Quick -> Weak Quick.

I also moved initialization methods to a `init()` function.


29988122:
Thank you for the help! Very appreciated it.
People from now on can have an Arts team thanks to you.

: DDD

PR #73, opened 2018-09-26
Adds a more modular preferred support selection.

TheIanator:
Had trouble getting the existing preferred support to work, lots of false positives and picking non-limit broken CE's, so I wrote this. It can't read as many parts of the support window as the previous function, so I made it a new one instead of editing the old one.

Can now mix and match servants and craft essences, as well as require a limit broken craft essence.

"SupportImageServantArray" can be an array of several servant portraits to check for, for example if all the people on your friend's list use different portraits for Waver. It can also be set to "Any", where it will not check for servants, just ce's.

"SupportImageCEArray" can be an array of several craft essences, for example if all your friends have different version of the +50 bond point CE. It can also be set to "Any" if only the servant is important.

"RequireLimitBrokenCE" when set to true will only select a support that has a limit broken CE.


potchy:
Hello, thank you very much for your time!
I had a quick look at the code and I think your ideas are on point.
Unfortunately, I won't be able to properly test it until next week, as I am very busy with work right now.
For now, I'll just drop a few points you can work on until then.

---

These ones **need** to be addressed before merging:
- You can replace the old function entirely. That's okay.
- At line 365, why did you choose to return a string instead of going the fallback route?

---

The ones below are just nitpicking. You can leave them to me, if you're not very experienced at refactoring.
- Lines 308-317 and 332-341 can be moved to a different function to reduce duplicated code.
- From line 302 to 348 there's a huge amount of nesting.

---

Finally, I feel like `SupportImageCEArray` could be a table instead so we would have a `RequireLimitBrokenCE` flag for each image. We can leave this one for later. What do you think?

TheIanator:
The line at 365 was just leftover from when I was testing and didn't want any actual clicks happening yet, will fix that and work on some of the other stuff soon.

modules/support.lua
```diff
@@ -0,0 +1,212 @@
...
+		return searchMethod.byServantAndCraftEssence
+		
+	elseif hasServants then
+		return searchMethod.byServant
+		
+	elseif hasCraftEssences then
+		return searchMethod.byCraftEssence
+		
+	else
+		scriptExit("When using \"preferred\" support selection mode, specify at least one Servant or Craft Essence.")
```
potchy: There's a new `scriptExit` call for when users leave both `Support_PreferredServants` and `Support_PreferredCEs` set to `"Any"`, as there is no point in doing so. It's better to just use the `first` option instead.

TheIanator:
Looks great. One more thing I was thinking of adding was detection for friends only, noticed that on both JP and NA there's a friend point icon in the lower right of the support window that is absent when the support is not a friend.

screenshot_20181002-174053
screenshot_20181002-173808

potchy:
@29988122 
Okay, I'm counting on you to update README.md then. I'm too lazy for this stuff. 😿
`SupportListRegion` has been restricted significantly.

screenshot_20180920-223126 1

@TheIanator
Nice find! I swear I never noticed that icon until now. lol
However, as that improvement is kinda out of the scope of this pull request, I'll just merge what's already finished and open a new issue for it instead.
For the time being, you can use the `Support_SwapsPerRefresh` option to somewhat limit the search area.

Issue #67, opened 2018-09-15
[Feature Request]Target Choose Should Happen in Each Stage


WoodenMug:
Currently the script will choose target once and never do it again because reset of TargetChoosen happen in menu stage: 
https://github.com/29988122/Fate-Grand-Order_Lua/blob/95c67f241196f846b6b3ac68dc2998c3f2cf6181/regular.lua#L756-L762
I suggest we move the variable out of the if clause (causing every battle action to check target) or putting it in the stage increment block (causing every stage to check target)

Thanks,
WoodenMug

29988122:
https://github.com/29988122/Fate-Grand-Order_Lua/issues/62
This one also.

WoodenMug:
Thanks, I can take a look at it and show you the snippet when I have something.

potchy:
Solved with #82

PR #82, opened 2018-10-14
#67 fix: When stage 2 had choosen target, Stage 3 won't choose target.

aqer:
1.rename CheckCurrentStage() and InitForCheckCurrentStage()
    to UpdateStageCounter() and InitForUpdateStageCounter()

    2.move UpdateStageCounter() from executeSkill() to battle(),
    executeSkill() use StageCounter directly

    3.if stage changes, reset TargetChoosen to 0, so every stage can choose
    target once.

### DIFF
--- a/regular.lua
+++ b/regular.lua
@@ -233,11 +233,13 @@ end
 
 function battle()
 	wait(2.5)
-	InitForCheckCurrentStage()
+	InitForUpdateStageCounter()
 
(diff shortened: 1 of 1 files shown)


29988122:
Yeah, I haven't thought of this solution <-really wanna refactor my bad codes.

Thank you!

PR #90, opened 2018-11-05
fix the script stuck  at result part when open  FastSkipDeadAnimation

Xianbei233:
and have tested  for 3 hours , works fine


Xianbei233:
update done.

Issue #114, opened 2018-12-04
Future of the script


TryBane:
I have been looking over the script and the Ankalua API a bit more in depth now that my schedule isn't as hectic as it was a few months ago. I think I can do an overhaul of the base script to allow for better scalability. I would move the psuedo-functional programming style to a more pure functional programming style rather than what we have now which is a procedural programming style with functional elements.

If this sounds too pretentious I understand as many have been working on this for a while and added some amazingly awesome features. But as the code base grows, the more complex and convoluted it becomes without a more solid foundation. I'm offering to rewrite the base code to be much more flexible and more error proof. Such as closing windows that were unintentionally opened. However, this will require help from many of you as I only have access to testing this with NA servers.

Once I have a working prototype, I can share it here for others to test with the other servers. What does everyone working on the code think?

TryBane:
For the record, some things I plan to include will be resolution scaling for different devices, better modularity for easier additions, and easier configuration for the user.

potchy:
Sure, I believe @ryuga93 is working on moving AutoSkill to a separate module. I'm also trying to improve code clarity in `battleLogic.lua`, but I didn't push anything yet because I'm not sure if something has broken in the process. lol

TryBane:
Ah, I see what you mean. I do think that separate scripts will help compartmentalize the workings to make it easier to make adjustments, but I was planning a complete rewrite from scratch. I'll at least get to work to keep the same functionality it has now without adding too many new features like error checking and such. For now, I'll just get it running so that it can do card selection. Maybe Autoskill.

TryBane:
Yeah, I know what you mean. Modules is along the same lines as what I mentioned with modularity. As we can freely add and remove certain elements of the script without worrying about ruining the running altogether. I'll do some work, and incorporate it into what you've worked on so far after you push the changes you've made.

This script is the only reason I've been farming so well during Christmas, so I thank you guys from the bottom of my heart!

29988122:
@anubi @TryBane @pokeguys
Thank you guys for trying to help us! Very much appreciated. 
As a side note - all the messy code were written by 2016 me.. : (

> How'd you evaluate other android UI interactive/scripting framework? 
I mean, maintaining the compatibility of an Android scripting framework required TREMENDOUS amount of work, and Ankulua did that pretty well. Not to mention that the author of Ankulua personally helped me & this project a lot. 

[...]

Issue #118, opened 2018-12-05
Script getting stuck on battle 1 after running for a couple hours


raspberrytealeaf:
I've been trying to babysit Nox while trying to see what causes it to get stuck on the drops screen (#115 ). Instead I found some other issue where it gets stuck on the 1st battle after running normally for a couple hours. [https://imgur.com/a/VmoDxv6](Here's a gif I made of the issue.)

Nox 6.2.6.0 and Ankulua Pro 8.0.5.

TryBane:
I am not sure if it is the cause, or if it is how it causes it, but removing the toasts made it work much longer for me. Also, I sometimes find that the script will not work right on the first iteration of combat, but then work fine for hours afterwards

29988122:
It looked like the script's in the routine: cast no skill & np, choose 3 cards only.

In this scenario, I'd personally close the extra window before the next toast "Battle 1/3", therefore the script can proceed normally.

I still suspect it's wait() related issue. Not really sure though.

We could make the script output debug.log but I think that's a little bit ... too much? Overkill? XD

TryBane:
No such thing as too much debug support XD
If we can get a log of what the script is doing during runtime it would help tremendously in solving issues that are happening.

29988122:
>TryBane

http://neopallium.github.io/lualogging/manual.html

This lib looked promising I guess? 
Anyway I'll edit, add this into our discussion thread.

29988122:
Check this closed PR for related discussion:
https://github.com/29988122/Fate-Grand-Order_Lua/pull/147

Issue #111, opened 2018-12-03
Anyway for script to avoid "Stunned" cards?


ParasiteEve:
Script always chooses Buster and Arts even if the client is stunned - just a note, thank you

potchy:
You've already opened an issue (#91) about this, but there's still no ETA for the fix. :confounded:
You can help by uploading a screenshot of the stunned cards, by the way.

ParasiteEve:
Sorry for the duplicate, I will add a screenshot the next opportunity I get.

Issue #40, opened 2018-05-23
Resolution problem


sliverlin:
你好我又來惹
這兩天換了新手機但是是18:9的解析度(2160x1080)
開啟FGO會出現左右的藍條，應該是這個緣故導致座標都亂了，腳本無法作用
想請問有沒有什麼方式可以校正
(目前我試了把腳本內所有起始X座標+120和按照2560&1920的比例+160都沒用)

![screenshot_2018-05-23-17-07-22-912_com komoe fgomycard](https://user-images.githubusercontent.com/39355740/40418153-2b90f8da-5eb4-11e8-9a8e-0a0c3343b033.png)

29988122:
請參照
https://github.com/29988122/Fate-Grand-Order_Lua/issues/33
Ankulua沒有處理的情況下，我真的不太好改

目前我的解法是請你用這個啦 XD
https://www.reddit.com/r/grandorder/comments/573lek/how_to_play_on_pc/

ryuga93:
Hi, maybe you can try to open the script with an IDE that can do refractoring or use Notepad++ to rename all the Region() and Location()

Write wrapper functions for Region and Location with the conversion and rename every instance with one click.

29988122:
We call that rewrite / refactor, right? I have to admit that currently I don't have a good coding style, so I'm afraid that any tweak of the script will break it and took too much hard effort to fix.

==
The function scriptdimension and comparedimension that Ankulua framework implemented caused this behavior. I checked the LineageM script written by Ankulua Dev, and frankly the solution seemed more complicated that what I've originally thought. We also have no idea whether they'll change the behavior about aspect ratio support or not, so I choose to wait.

29988122:
真虧你想得到這種特殊解.....不過18:9的手機不只有一隻，什麼s8s9小米zf5之類的，一卡車

為了日後寫程式的相容性我還是只能等Ankulua官方做處理，特別寫例外狀況會很坑 囧


總之恭喜你找到解！ XD

ryuga93:
I would say it's the laziness of game dev to support 18:9 that caused this problem XD

pokeguys:
Using OpenCV can easily fix this issue. But it required an external API server to process and communicate between the PC and App due to Ankulua doesn't provide any OpenCV function in Lua. 🤦‍♂️ If you need this feature, I can create a PR (I am not quite sure it is 100% work. I don't have any device > 16:9 resolution)

## Result
[...]

PR #121, opened 2018-12-06
Major refactoring

potchy:
Hi, guys. I finally finished doing my changes, but I'm feeling conflicted about some decisions, so I thought it was better to open a pull request so we can review them together before merging everything to `master`.

# Module splitting
First, I'll briefly explain my thought process when deciding the responsibilities of each module.
There are some issues down the line, but let's focus first on the primary changes I made.

- **autoskill.lua**: mostly implemented by @ryuga93, its only responsibility is to click skills according to user preference. It used to also contain logic to update the stage counter and the like, but I think they fit better in `battle.lua`, so I moved them there.
- **card.lua**: holds everything related to cards: where they are located on-screen and how to classify and priorize them by affinity (Weak, Normal, Resist) and type (Buster, Arts, Quick), also according to user preference.
- **battle.lua**: holds generic battle logic, like in which stage/turn the battle is, what targets to prioritize, how to enter the card selection screen. It orchestrates `card.lua` and `autoskill.lua`'s actions as well.

# Naming convention for local module variables
@TryBane made [a pretty good point](https://github.com/29988122/Fate-Grand-Order_Lua/issues/114#issuecomment-444252560) about local module variables not needing an underscore to differentiate them, but what happens when we do need to expose them?

In `battle.lua`, there is a field called [_hasClickedAttack](https://github.com/29988122/Fate-Grand-Order_Lua/blob/75ed8124570a900695c21b7b10048c14c903d369/modules/battle.lua#L29) and a function called [hasClickedAttack()](https://github.com/29988122/Fate-Grand-Order_Lua/blob/75ed8124570a900695c21b7b10048c14c903d369/modules/battle.lua#L208) which exposes that field to the outside world. Without the underscore, both the field and the function would have the same name.

That wouldn't be an issue if we used **camelCase** for fields and **PascalCase** for functions, but I think we should really stick with **camelCase** for functions instead, since AnkuLua already follows this convention.

# Coding style regarding impure functions
You probably know already what "impure" and "pure" functions are, but let's just clarify it, so we're on the same page.

An **impure** function is a function with side effects. That is, it changes something outside of its scope. In the context of this script, a simple example would be [battleLogic.lua:ChooseTarget()](https://github.com/29988122/Fate-Grand-Order_Lua/blob/4c0a8d5c82f5ba10b85e863ee38ba563b64f4ce5/modules/battleLogic.lua#L151). Calling it will change `TargetChoosen`'s value to `1`, so in the next turn of the battle, the script will know it doesn't need to choose a target again, because it already did that previously.

A **pure** function doesn't change anything outside of its scope and its 100% certain that, for a given input X, it will produce Y effect and return Z. That's why it's called **pure**. A good example would be [battleLogic.lua:checkCardType(region)](https://github.com/29988122/Fate-Grand-Order_Lua/blob/4c0a8d5c82f5ba10b85e863ee38ba563b64f4ce5/modules/battleLogic.lua#L191). Considering that in `region` there is an image of a Buster card, it's 100% certain that this function will return `BCard`. Not `ACard` or `QCard`, but `BCard`. Always.

Are impure functions evil? Not really, but they can be a little bit harder to debug and understand. Anyway, back to the point, if you take a close look at the new [battle.lua](https://github.com/29988122/Fate-Grand-Order_Lua/blob/75ed8124570a900695c21b7b10048c14c903d369/modules/battle.lua), you'll notice there are some new functions prefixed with "on". This prefix explicitly declares that those functions have side effects. How so? Look at the function below.

```lua
chooseTarget = function(targetIndex)
	click(TARGET_CLICK_ARRAY[targetIndex])
	onTargetChosen()
end
```
This function says "hey, I can click on this target thing for you, but by doing so, something else is going to happen". This "something else" is defined in `onTargetChosen()`.

```lua
onTargetChosen = function()
	_hasChosenTarget = true
end
```

I believe it's good practice to have names like that, because it's much easier to spot which functions are the culprits behind side effects and to understand what **event** triggers them, in the first place.

# Structural problems
And here we are. *inhales*

What I wanted to accomplish with this refactoring was to have parent-child relationships between modules. I wanted `regular.lua` to depend on `battle.lua`, which in its turn would coordinate `card.lua` and `autoskill.lua`. Here's a visual representation:

fgo lua-what i wanted

But... I noticed that wasn't gonna happen. Here are the reasons:
- `autoskill.lua` depends on `battle.lua` to know in which stage the battle is and if the script already clicked the "Attack" button and click it, if necessary, for casting Noble Phantasms.
- `autoskill.lua` depends on `card.lua` to know where the NP cards are located on screen, so it can click them.
- `battle.lua` depends on `autoskill.lua` so it can can call `executeSkill()` at every turn of the battle. 
- `battle.lua` depends on `card.lua` to know in which cards to click.
- `card.lua` depends on `autoskill.lua` to know if all NP commands have already been processed, so it can decide if NPs are to be cast or not, when `Battle_NoblePhantasm` is set to `"spam"`.
- `card.lua` depends on `battle.lua` to know if a priority target has already been chosen, so it can decide if NPs are to be cast or not, when `Battle_NoblePhantasm` is set to `"danger"`.

Finally, I ended up with this enormous mess where everything depends on everything:

fgo lua-what i got

Not only that, I learned that it's not possible to use `require(module)` to link modules this way, because Lua can't handle circular dependencies. If `card.lua` requires `battle.lua` and `battle.lua` requires `card.lua`, there's no way for the Lua interpreter to know which to load first.

circular-dependencies

So, I had to glue them together this way:

```lua
regular.lua:
local card = require("card")
local battle = require("battle")
local autoskill = require("autoskill")

function init()
	autoskill.init(battle, card)
	battle.init(autoskill, card)
	card.init(autoskill, battle)
end
```

It works, but it looks terrible and confusing.
I'm not criticizing the language, as I know this is bad design, so I'm thinking about splitting things into more modules, but I'm not sure how, if even feasible.

# Practical problems
Bad things aside, here's an easier problem for us to work on.
I managed to significantly speed up the script, as you can see in [this video](https://www.youtube.com/watch?v=RQiYGpbjV9M), featuring my Shiki slaughtering bad dragons on Christmas™.

However, at **00:36**, **00:53** and **01:10**, you'll see that the [card.getCardType(region)](https://github.com/29988122/Fate-Grand-Order_Lua/blob/75ed8124570a900695c21b7b10048c14c903d369/modules/card.lua#L143) function fails and prints an error message on screen. I assume it's because of the "No priority target selected" message covering the region where the 3rd command card is. 

after_refactoring mp4_snapshot_00 36 333
after_refactoring mp4_snapshot_00 53 333
after_refactoring mp4_snapshot_01 10 000

The result is especially bad in low DPI devices, where that message covers one more card:
screenshot_20181205-224105

Therefore, we have two options to solve this problem: either we get rid of the `toast(message)` or we put a `wait(time)` long enough for it to disappear before the script checks for the card types.

Additionally, at **01:43** and **03:04**, the [card.getCardType(region)](https://github.com/29988122/Fate-Grand-Order_Lua/blob/75ed8124570a900695c21b7b10048c14c903d369/modules/card.lua#L143) function fails again when checking the 5th card, even when there wasn't any messages covering it.
after_refactoring mp4_snapshot_01 43 167
after_refactoring mp4_snapshot_03 04 167

I'm still not sure why, but I think expanding [these regions](https://github.com/29988122/Fate-Grand-Order_Lua/blob/75ed8124570a900695c21b7b10048c14c903d369/modules/card.lua#L17) a little would help.
card_type_regions


TryBane:
Regarding toast messages, I think it would be best to use those for debugging or just purely for displaying error messages. I don't see a need to display when battle stage is presently on, except for debugging which doesn't need to be in the release version.

As for the dependencies, that can be solved either with creating a dependency script which holds all of the shared factors like variables and regions, or by doubling up on them. Such as providing a region for the NPs in both AutoSkill.lua and Card.lua.

And regarding the camelCase for functions, I think we can avoid that by creating wrapper functions for the AnkaLua API functions themselves with some debugging added for the non-release versions.

ryuga93:
I'll start with the practical issue, since I'm seriously weak in code design and styling.

If we could move from from toast to something like logging as mention in https://github.com/29988122/Fate-Grand-Order_Lua/issues/118#issuecomment-444486172, it will be better I think. And we can choose when to log with `isDeveloper` flag.

That aside, I'll look into the 5th card issue.

ryuga93:
@potchy changing the last region from `Region(2280,1060,200,200)` to `Region(2180,960,350,350)` should fix the problem. 

Not sure about the other server though. Need @29988122 to verify.

ryuga93:
`In battle.lua, there is a field called _hasClickedAttack and a function called hasClickedAttack() which exposes that field to the outside world. Without the underscore, both the field and the function would have the same name.`

We can make the getter functions to start with get followed by field name, eg: getHasClickedAttack()

TryBane:
Yeah, getter and setter functions are fairly standard. Though setter
functions tend to be shied away from when you want strict encapsulation. I
don't think this is large enough of a project to be TOO worried about
strict encapsulation though so I don't think it would be a huge issue.

29988122:
Feel free to merge, too busy right now..

BTW JP server will add a filter to show MLB equipment only friend.

Gotta add that into sup choosing @potchy

potchy:
> @TryBane
> Regarding toast messages, I think it would be best to use those for debugging or just purely for displaying error messages.

I agree.
We should just get rid of them~

> And regarding the camelCase for functions, I think we can avoid that by creating wrapper functions for the AnkaLua API functions themselves with some debugging added for the non-release versions.

That's a tad overkill for naming purposes. lol

---

> @ryuga93
> changing the last region from Region(2280,1060,200,200) to Region(2180,960,350,350) should fix the problem.

Yup, that worked for me.

> We can make the getter functions to start with get followed by field name, eg: getHasClickedAttack()

While I think this is fine for stuff like `getName()`, it doesn't suit booleans at all. `getHasClickedAttack()` or `getIsSomething()` looks idiomatically wrong. In that case, I prefer to just use **camelCase** for fields and **PascalCase** for functions then.

> Just like @TryBane mentioned, we could borrow the Redux concept, extract the states out and place them in a "database", then use "reducers" to perform "actions" on the "database". Regular.lua can act as the conductor for the orchestra of modules.

I've done by bit of research on this pattern and I think it suits this project very well. If I understand correctly, `regular.lua` would create a "battle" store and pass it to `autoskill.lua`, `battle.lua` and `card.lua`. To issue state changes, they would "dispatch" actions to the store, which would redirect them to the corresponding reducers. Did I get that right? If yes, would these reducers be able to call functions in `battle.lua`, for clicking the attack button, for example? I got a bit confused in that part.

---

```
@29988122
BTW JP server will add a filter to show MLB equipment only friend.
```

But I don't even have a JP account, mister. 😖

TryBane:
I am unfamiliar with Javascript and the term Redux, but that design pattern sounds about what I was considering as a solution.

Using this project as an example, I am used to programming in such a way that you could just require a region passed to AutoSkill's ExecuteSkill() function, or a list of regions at least, which would then be acted on within the script itself. Either that, or you just create a field which is then populated in an init() function called within regular.lua. The latter method would prevent constant creation and destruction of locals but at the cost of maintaining that space over the course of the script.

I believe the latter would be best for performance reasons, but I am unsure of how well Android OS's can handle long-term memory storage. To give a more in-depth example:

AutoSkill.lua
`
        local skillRegions = {}
        function init(theSkillRegions)
                skillRegions = theSkillRegions
        end
`
regular.lua
`
        local AutoSkill = require("modules.AutoSkill.lua")
        local skillRegions = {Region(1234,1234,12,12),Region(1234,1234,12,12)}
        function init()
        AutoSkill.init(skillRegions)
`

This is of course a simplified example, but I'm sure you can get the gist of it.

Also, if we are creating Arrays, are we starting from index 0? Cause that is pretty much what I default to XD

potchy:
@TryBane 
Actually, when you require a module, Lua stores a single reference to it somewhere, so it is not loaded twice even if you require the same module in two different files. As there's no additional memory usage in that regard, we don't really need to worry about it.

And about array indexes starting from 1, "when in Rome, do as the romans do", so I reject your idea. :triumph:
<sub><sup>\*secretly thinks every language should start indexes from 1\*</sup></sub>

---

@ryuga93 
Oh, I get. The Observer pattern is basically the same thing as listeners in Java Swing, Android and events in .NET.

But then, Redux may not fit this project at all. Looking at [this](https://github.com/zebMcCorkle/lua-redux/blob/master/lua-redux.lua), I assume there's only one "subscribable" event and every single state change will notify all subscribers. So, even if battle.lua is waiting for changes only in the `hasClickedAttack` field, if someone else changes `isOrderChanging`, `battle.lua` will still receive a notification that something has changed. We could probably go with something like [Redux Observable](https://codeburst.io/redux-observable-to-the-rescue-b27f07406cf2), which is a mix of both worlds. Then, we would have a state store that also notifies subscribers of changes in specific fields.

Anyway, this may need some studying.

---

I'm considering on merging this as is, so we can apply the new naming conventions to all files and then @TryBane can proceed with the changes he suggested in #111. Everyone agrees?

ryuga93:
I agree, Redux alone won't fit the project.

@TryBane I have merge the pull request. 

And I think it is time to freeze the code base. Only minor changes that fixes broken functionality are allow, as well as documentation updates. I'll also draft a final release for version 2, since version 3 will focus on structure renewing. Then further functionality  improvement can start on version 4.

29988122:
@ryuga93 
>Not sure about the other servers though. Need @29988122 to verify.
Worked in JP version, too. Not tested for long though.

@potchy
>*secretly thinks every language should start indexes from 1*
LUL.

>Observer pattern, or reducer, action, database
Read all of the discussions. 
So that's why I have a listener class in my telegram python bot..I just copied and pasted the examples. xD
I have already read some design pattern books/articles, but, still, only when you encountered it you wish you'd known better.

So, currently: 
1. Waiting for @TryBane to draft his improvements / PR, right?
2. I'll redo the readme.md for *users*, leave all the details for github dev pages.
3. And as @ryuga93 suggested, we shall start to write the documentation on our *Grand Order* github pages.

AMIRITE?

TryBane:
As far as I'm aware, yes. I am currently working on the redraft. Gotta do some testing to ensure everything works correctly. But aside from that, should not take long to make the actual changes after everything is confirmed to work.

Currently working on setting up the scaling of the button clicks and regions based on the resolution. After I get everything understood, I shouldn't have any issues.

29988122:
Ehh, sorry to inform, but as I eat my 3rd apple, the script (finally?) failed to detect 5th card.

We should ignore it tho that as we shall wait until the fundmantal's done.

TryBane:
Quick update:

I've converted all Regions and Locations to be based on getAppUsableScreenSize():getX() and getY() functions. There are still some bugs to work out regarding it. but I can post what I've got for a test on other devices once it is finished.

Also, regarding the Pull Request here #124 I have also noticed and fixed this as well. It was causing the Skill_Command that is listed in the FGO_XX_REGULAR.lua to be used rather than whatever was being written in the List. This was because the init() functions were being called BEFORE the List was being parsed for the one the user wanted. So I've solved this as well but I don't see any reason for someone else to solve the issue themselves. I think I can also solve the issue regarding the card at the end not working as well. We will see. First need to solve all the bugs that came out from moving from hardcoded resolution numbers to scaling.

Also, I want to just inform everyone that my method of using those values mentioned above is as such. I create 2 const values APP_WIDTH and APP_HEIGHT and then set them equal to their relative functions.

Then in the Regions and Locations I use them as such, but this is only an example:
```lua
APP_WIDTH = getAppUsableScreenSize():getX()
APP_HEIGHT = getAppUsableScreenSize():getY()

Region(APP_WIDTH*.62,APP_HEIGHT*.235,APP_WIDTH*.037,APPHEIGHT*.012)
```

potchy:
I didn't get why you changed those values.
AnkuLua automatically scales everything, so why? Unless we're talking about the borders in #40

TryBane:
The hope is that it will allow for multiple aspect rations. As mentioned, I'd have to test on devices such as tablets but if it works then it would be much more flexible than the current method. Should it have issues, then we can always do away with the changes and I can concentrate solely on the full reactor.

My goal is to first ensure that everything works as is currently with the new changes before moving to the full reactor. Should be completed tonight then I can begin testing.

potchy:
Oh, here's a checklist of weird resolutions:
- 1280x720 (16:9)
- 2048x1536 (4:3)
- 1792x828 (448:207)
- 812x375 (812:375)
- 2688x1242 (207:448)
- 2388x1668 (199:139)
- 2732x2048 (683:512)
- 2160x1080 (2:1)
- 2246x1080 (1123:540)

I'm also working on a solution for this problem, since now it's of my interest as well.
I've bought a Xiaomi Redmi Note 5 (2:1 AR). :pensive:

29988122:
Report on new support selection system UI:

1. You can filter to show only certain EVENT CE(s), therefore less recognition error and less list refresh. Our support selection module is not affected.

default

(Yes, that orange one is a button. very bad design choice)

2. You can filter to show all MLB CE(s) only, not only event MLB CE(s).
Our support selection module is not affected.
default

3. The "Mix" class sorting tab shows all the servants/friends AT ONCE.
That is a fucking long page(10+ swipes).

default

That is, the whole "Mix list" of your friends+non-friends servant is exactly what shows in other 9 tabs combined. When the UI change applies to other servers, users should only use this page for CE only support servant selection, unless they want to specify servant - then the class specific page should be used.

All in all, a great improment for JP player, although the UX can't be messier.

potchy:
I'm not surprised the UI is getting more cumbersome as time passes. I find that's pretty common for online games. They try to impress players with new design, but don't bother to rework the older stuff. Granted, Grand Chase and Elsword were the only games I played "seriously".

29988122:
Because they did not hire a proper mobile UI/UX designer, nor did Japanese have this kind of people. (AND YES I MEAN IT.)

Unless it's a console game, or it's in the planning stage of development.

ryuga93:
The only impressive part I find about most mobile games was the amount of effort they put into the anime like trailers.

TryBane:
Just an update on my progress, I've finished the adjustments to the resolution scaling. Everything works on my phone. I'm going to be testing it on multiple resolutions to see if it still works as expected. Sorry for the delay!

After I have confirmed that everything works as it should, I will proceed to do the complete overhaul! My goal is to get it done before Solomon in NA arrives so that people can take advantage of a faster farm for Barbatos and the other Pillars.

TryBane:
Hey guys, finished with adding in resolution scaling. I hope this works on multiple devices, but I can't seem to get it to work on Nox. I don't know what I'm doing wrong. @potchy Would you being able to test it on there for me?

A few things I changed aside from resolution scaling was screen recognition for skill casting so that skills are cast as soon as they are able to be cast, I also utilized the continueClick() method rather than the click() method for tapping the screen at the Result window in order to quickly move past the results. I need an image (to scale) of a bond CE being acquired in order to get the image and region for recognition to work in order to skip past it and such, so as of this moment that isn't supported (And I think I'll be running into that issue come Solomon)

Also suggest for anyone using this to use fastclick if possible, it actually does the entire farming faster than I do! And that's assuming I'm clicking as fast as possible already knowing what I need to click!

As a note, my fastclick settings are 50.0 milliseconds down, 50.0 milliseconds delay.

I also added an additional card click to the card selection because there was a bug (possibly from using fastclick) where on turn 4 when I don't have an NP to cast, it would skip a card and only select 2 causing it to get stuck at the card selection window. As I'm unsure of the reason why, I made a simple solution to at least keep it from getting stuck in those instances and it seems to work now.

[Lua Script.zip](https://github.com/29988122/Fate-Grand-Order_Lua/files/2700196/Lua.Script.zip)

29988122:
https://help.github.com/articles/creating-a-pull-request/

Try this maybe? You can draft a PR now by following this tutorial(or other tutorial find online)

By doing this way we're more capable to see all the changes and diff+-.


Thank you bro!

TryBane:
I'm not able to make a PR for this github project, so I cloned it to my own version, branched it, and made the changes in a commit. Then I made a PR on my own version. I'll link the PR here for you guys to review.

https://github.com/TryBane/FGO_LuaAdjustments/pull/2

potchy:
Okay, I tested your code in two different resolutions: one with borders on left/right and one with borders on top/bottom. It didn't work in neither.

Here's where `MenuRegion` ended up being in 2048x1536:
screenshot_2018-12-21-12-12-52

And in 2688x1242:
screenshot_2018-12-21-12-17-42

It's not a complete loss though, since this gave me an idea.
First, we'll move all locations/regions to a different module where they will all be stored in a single table. Then, we'll calculate the size of the game's borders, according to the screen resolution.
Thereby, it's just a matter of iterating that table and appending an offset (X, Y) to every location/region.
I'm fairly knowledgeable in how image scaling algorithms work, so leave this to me!

Regardless, I liked all of the other changes, so please open another pull request without the resolution changes. This repository accepts public pull requests, so there's no need to open one in your own fork.

PR #124, opened 2018-12-10
fix: Autoskill_List doesn't work

aqer:
Bug: init Autoskill_List commands after insert command table,
thus autoskill always use Skill_Command.

By the way, I think Skill_Command can be replaced by Autoskill_List.
It's a bit strange to have two similar skill system.


potchy:
thanks

Issue #21, opened 2018-01-12
Discussion with my collaborator


29988122:
Having a new discussion thread seemed like the fastest way for us to discuss non-commit and non-PR related issue. @ryuga93 

I can't figure out the function calling here:

```
function menu()
    atkround = 1
    npClicked = 0
    turnCounter = {0, 0, 0, 0, 0}
    click(Location(1900,400))
    wait(1.5)
    if Refill_or_Not == 1 and stoneused < How_Many then
        refillstamina()
    end
    click(Location(1900,500))
    wait(1.5)
    click(Location(2400,1350))
	wait(8)
end

function refillstamina()
    if StaminaRegion:exists("stamina.png", 0) then
        if Use_Stone == 1 then
            click(StoneClick)
            click(Location(1650,1120))
            stoneused = stoneused + 1
        else
            click(AppleClick)
            click(Location(1650,1120))
            stoneused = stoneused + 1
        end
    wait(1.5)
    menu()
    end
end
```
The 3rd line from last memu() called inside refillstamina() was the one you recently added.
Wouldn't it caused a lot of wait(1.5)? 

1.Every time when refillstamina() was called but game's not currently at AP refill window.
2.Every time memu() was called).

Or I got the whole function calling inside function concept wrong?

ryuga93:
The wait in refillstamina() will not be called if stamina.png does not exist.

I think I forgot to indent both functions...

29988122:
sure I understand, but then it will keep calling this part
```
click(Location(1900,400))
 if Refill_or_Not == 1 and stoneused < How_Many then
        refillstamina()
    end
```
And misbehave, right? Just tested that.

If I did the permission right, you should be able to commit freely now(collaborator thing).
Make changes as you wish! : )

ryuga93:
Once it click on the confirm and refilled the stamina, the game will return to the home window, at least for EN.

I change this because for me the script lack one step and stop at friend selection window.

Will it move to friend selection window after refill for JP?

29988122:
**Yes it will**

Wow, version differences.......
We should make an extra variable to bypass this issue.

I guess only in JP it will move to friend selection directly, which is probably NOT the case in other countries. Because I played the game since 1st day, so I remembered this behavior change.

29988122:
https://github.com/29988122/Fate-Grand-Order_Lua/commit/ca56a42245ec4f18398d1c65c7cbc918e5ed99fa
[...]

PR #139, opened 2018-12-22
Took a bit of tinkering

TryBane:
But I managed to get it.

### Changes in autoskill.lua

This will allow for image recognition to be used rather than straight wait times during farming. @potchy made it so that the image being searched for is a better option so that extensive wait times are uneeded during Raid/Boss fights and still functions almost the same as normal. There is a slight delay in comparison due to the screen capture of the stage taking a couple seconds (which was handled by the time skills could be clicked) but it is better than using an arbitrary number for a wait time that may change in any number of situations.

### Changes in regular.lua and FGO_NA_REGULAR.lua

Also added the functionality of choosing multiple types of apples along with All Apples allowing you to clear out your current apple stock in order from Bronze->Silver->Gold. Changed "How_Many" to "Repetitions" to reflect the change since Bronze Apples can be used in multiples. Also added comments to explain this functionalities use. Also please note that I have NOT made these changes in the other FGO_XX_REGULAR.lua files! Someone please get that for me? I don't know how to write the explanations and such in those languages ;(

### Changes in card.lua

Changed the range for the search regarding the command cards (fixes the bug someone pointed out previously) Also changed the required 3 card clicks to 5 cards to account for the occasional issue where someone uses fastclick and the script skips a card.

### Small notes added to autoskill.lua

I made a note regarding the variable "isFirstSkill". With the use of the new image for recognition and using image recognition itself for determining when to cast skills, I do not see the point of having this variable any longer. I left it in in case there was something I was unaware of that it affected, but thought I'd point it out. The only use I saw was in decodeSkill() where it ensures you have enough time for skill animations.


potchy:
I'll only have time to review the PR thoroughly by tomorrow night, but I'll leave a few suggestions for now.
It's currently 1am here. lol

modules/autoskill.lua
```diff
@@ -148,9 +149,16 @@ executeSkill = function ()
 		local isFirstSkill = 1
 		if currentSkill ~= '0' and currentSkill ~= '#' then
 			for command in string.gmatch(currentSkill, ".") do
+				-- Check that skills can be clicked
+				while not _castingNP and not BATTLE_REGION:exists(GeneralImagePath .. "battle.png" ) do end
```
potchy: I think `_castingNP` can be replaced for `_battle.hasClickedAttack()`.
And the infinite loop can be replaced for a `wait("blablabla.png", 10)`, where 10 is a timeout. Maybe just 5 seconds is enough?

regular.lua
```diff
@@ -105,29 +122,11 @@ function startQuest()
...
+	
+	-- Click through all of the Result screen items. Does NOT account for Max Bond CE acquisition (yet)
+	continueClick(QuestResultNextClick,35)
 
-	--Extra event item screen.
-	if isEvent == 1 then
-		wait(1.5)
-		click(QuestResultNextClick)
-	end
+	wait(3)
```
potchy: There should be enough delay in line 127, so I guess lines 129 and 132 can be safely removed.

regular.lua
```diff
@@ -226,7 +229,7 @@ while(1) do
 	if battle.isIdle() then
 		battle.performBattle()
 	end
-	if ResultRegion:exists(GeneralImagePath .. "result.png", 0) then
+	if ResultRegion:exists(GeneralImagePath .. "result.png", 0) or BondRegion:exists(GeneralImagePath .. "bond.png", 0) then
```
potchy: Does `BondRegion` overlap `ResultRegion`? I can't recall...

modules/card.lua
```diff
@@ -32,7 +32,7 @@ local COMMAND_CARD_CLICK_ARRAY = {
 }
 
 local NP_CARD_CLICK_ARRAY = {
-	Location(1000,220),
+	Location(1000,400),
```
potchy: I believe @29988122 used `220` instead of `400` for a reason.
My guess is that it prevents the script from accidentally changing targets, because the 1st NP card slightly overlaps the region where the rightmost enemy is.

He should have left a comment about it, though. :sweat_smile:

modules/autoskill.lua
```diff
@@ -148,9 +149,16 @@ executeSkill = function ()
 		local isFirstSkill = 1
 		if currentSkill ~= '0' and currentSkill ~= '#' then
 			for command in string.gmatch(currentSkill, ".") do
+				-- Check that skills can be clicked
+				while not _castingNP and not BATTLE_REGION:exists(GeneralImagePath .. "battle.png" ) do end
+				
+				-- With the use of screen recognition, I do not believe it is necessary to use isFirstSkill anymore /TryBane
```
potchy: Okay, so please do remove it. o_ó

modules/autoskill.lua
```diff
@@ -148,9 +149,16 @@ executeSkill = function ()
...
 		if currentSkill ~= '0' and currentSkill ~= '#' then
 			for command in string.gmatch(currentSkill, ".") do
+				-- Check that skills can be clicked
+				while not _castingNP and not BATTLE_REGION:exists(GeneralImagePath .. "battle.png" ) do end
+				
+				-- With the use of screen recognition, I do not believe it is necessary to use isFirstSkill anymore /TryBane
 				decodeSkill(command, isFirstSkill)
 				isFirstSkill = 0
+				
+				wait(.2)
```
potchy: why

potchy:
By the way, when you push more commits to your branch, they will be automatically added to this pull request.

TryBane:
I agree that hasClickedAttack can replace the variable I used there. Didn't think of that.

I used the infinite loop because using wait requires you to adjust the frequency of scans otherwise it'll almost always hit the delay. However, I don't want there to be an instance where a servant is later released with a longer animation than the current longest one. Also, I felt this would allow clicking on targets to be quicker than waiting for a timeout.

Bond region does not overlap with the result screen. This is one of the reasons that the script was hanging up on bond level ups as stated in a few issues.

I didn't want to remove the isFirstSkill if it was used elsewhere. But if you think it can be safely removed then I will do so.

And the reason for the .2 wait time is because it provides enough delay for the screen to remove the battle.png image and enough time for the target select screen to come up for skills.

And thank you for the information about further commits. that keeps me from having to make a new PR XD

ryuga93:
```
local SKILL_CLICK_ARRAY = {
	[  1] = SKILL_1_CLICK,
	[  2] = SKILL_2_CLICK,
	[  3] = SKILL_3_CLICK,
	[  4] = SKILL_4_CLICK,
	[  5] = SKILL_5_CLICK,
	[  6] = SKILL_6_CLICK,
	[  7] = SKILL_7_CLICK,
	[  8] = SKILL_8_CLICK,
	[  9] = SKILL_9_CLICK,
	[ 10] = MASTER_SKILL_1_CLICK,
	[ 11] = MASTER_SKILL_2_CLICK,
	[ 12] = MASTER_SKILL_3_CLICK,
	[-47] = SERVANT_1_CLICK,
	[-46] = SERVANT_2_CLICK,
	[-45] = SERVANT_3_CLICK
}
```
You can replace the numbers with chars, and expand the logic from range conditions to equality, but it will be much longer to write. Performance wise I think it will go unnoticeable.

TryBane:
You can still use range conditions with chars, should not have much impact on difficult to code.

ryuga93:
Is it? My whole programming life was a lie OTL. I had never used > or < for chars. But I think it makes sense since char is just uint8 lol

TryBane:
Yeah, in my own version on my phone where I am using isCastingNP, I didn't know which set of conditions were for NPs without reading the documentation (fixed that for my commits) I used > and < and checked the char instead of the index.

TryBane:
Adjustments made. I can't add comments on mobile for some reason. Otherwise I'd label each section about what it does.

TryBane:
I just got home. So I added some comments to make it easier to understand and caught a bug that might've been bad. That was fixed without issues and hopefully the code makes more sense to would-be readers who are not already familiar with what the code itself does.

I am going to Create a branch off of this and work on that for now to begin the true Refactoring that we talked about extensively. This was mostly a feature overhaul for the autoskill and refill features, along with a couple of bug fixes.

TryBane:
I have made multiple fixes to some issues that plagued my initial draft. I have tested it and do believe it works as expected though I will be adding another slight delay to right after it finds the battle.png because I believe the order of fade in (bringing the interface up after skill/battle animations) prioritizes the new image first, meaning the skills might get clicked a tad too early when they haven't had the opportunity to become available to click.

To note a few of the issues I saw, I had changed all of the indexes of the autoskill.lua file to be based on chars, and the clicks all used str, but I forgot to change the 12 to "l" for that particular instance when you order change. Also still had an instance of index in there, so fixed that as well.

There was another one where I added in AcceptClick as a place for clicking, but never actually initialized it as a variable.

From what I recall those were the only issues I saw.

FGO_EN_REGULAR.lua
```diff
@@ -23,8 +23,19 @@ end
...
+    "All Apples"    Use all available Apples for refills
+    "Gold"          Use only Gold Apples for refills
+    "Silver"        Use only Silver Apples for refills
+    "Bronze"        Use only Bronze Apples for refills
+]]
+Use = "All Apples"
+
+-- Represents the amount of times a refill will happen.
+-- Is NOT a counter for number of Apples used when Bronze apples are included
+Repetitions = 0
```
potchy: I think the refill parameters' names should be changed to fit our new convention, like so:
```lua
Refill_Enabled = 0
Refill_Resource = "SQ" -- also change Stone to SQ to conform with the official localization
Refill_Repetitions = 0
```

modules/card.lua
```diff
@@ -198,8 +198,9 @@ clickCommandCards = function()
 		for _, cardSlot in pairs(currentCardTypeStorage) do
 			click(COMMAND_CARD_CLICK_ARRAY[cardSlot])
 			clickCount = clickCount + 1
-	
-			if clickCount == 3 then
+			
+			-- Increased clickCount requirement from 3 to 5 to prevent fastclick from missing a click
+			if clickCount == 5 then
```
potchy: Regardless of this condition, if the script misses a click, your card order will be messed up anyway, so isn't it better to increase your fast click delay a bit instead?

modules/autoskill.lua
```diff
@@ -141,15 +142,15 @@ executeSkill = function ()
...
-			isFirstSkill = 1 means more delay, cause one has to wait for battle animation.
-			isFirstSkill = 0 means less delay.
-		--]]
-		local isFirstSkill = 1
 		if currentSkill ~= '0' and currentSkill ~= '#' then
 			for command in string.gmatch(currentSkill, ".") do
-				decodeSkill(command, isFirstSkill)
-				isFirstSkill = 0
+				-- Check that skills can be clicked
+				while not _battle.hasClickedAttack() and not _battle.isIdle() do end
```
potchy: This line really improved AutoSkill performance.
My device takes around 15 seconds to process Order Change, so that old `wait()` call just wasn't cutting it. lol

potchy:
5th Buster Card bug still lives
JESUS CHRIST WHY

screenshot_20181222-150505

TryBane:
I could've sworn I made changes that would've fixed that. But oh well, you got it so that's fine.

Is there anything else you can think of that can be changed? I just woke up and am ready for some programming :D

TryBane:
Ah, you're right! My bad. Also, might wanna make a small adjustment regarding the wait times for clicking on the Bronze Apples as the toast gets in the way.

TryBane:
To be specific, in your example you only have 1 apple left and currently only have 29 stamina. In that instance, it the bronze apple option will be grey'd out even with an apple there as it won't be enough to open the quest.

The only instance where you can choose bronze apples regardless is when you enter the refill screen manually instead of from attempting to enter a quest you don't have AP for. Btw, AP is the shorthand used in NA, standing for Action Points. Anyways, I personally had this situation where 1 bronze apple was left to me but wasn't enough to refill the bar. And it didn't get stuck.

So unless they changed it in JP, which is entirely possible, then it shouldn't cause an issue. What I would say you should do is if you have a single bronze apple and have less than 40 Stamina, attempt the daily xp and see if you can click the bronze apple. That will enable us to be sure whether it works or not. Otherwise we can perform emergency fix for it should it be an issue and we can't test for it.

29988122:
OK it is 120% different client UI behavior issue. Happened multiple times in the past already.

Sooner or later as the client in other servers will catch up, and adapt those new UI behavior.
I'll figure out a way to deal with them more elegantly.

TryBane:
My solution will have to expand in that instance. We can use the region to branch the behaviour into the current or a different version that can handle the situation (after pressing okay, we check if the stamina.png is still there and repeat the process.)

I will take care of this now as I literally just pulled into my driveway XD

TryBane:
Oh, and if you know, could you give me all the versions that follow this behavior? I assume CN and JP do, not sure about TW, and I assume KR is like NA in this instance seeing as they are around the same area in the story.

29988122:
Check the video. It's a little bit different from what I described(memory's unreliable), but the point still stands.
https://youtu.be/hxjE3_ULnck
So I'll leave that to you(autorefill tweaking), thanks bro!(I only have enough time working on readme at the moment tho)

***

Regarding "version differences", according to this comment's date:
https://github.com/29988122/Fate-Grand-Order_Lua/issues/21#issuecomment-357257089
Change happened around Jan 2018 or late 2017.
 
That is, AFTER event "Christmas event for Ereshkigal", but BEFORE event "Climbing tower with Shuten Doji"

Of course, the exact version / date is unknown.
So I suggest ....... we ignore other versions, unless people complain about it(yeah, I'm being serious, cause we only play JP/EN server, thus we can only maintain them.)

TryBane:
I made a fix for the issue you mentioned here already in another pull request. The behavior of this fix is it checks if it is "EN" based on the GameRegion variable we use in each of the FGO_XX_REGULAR.lua files, and if is "EN" it follows the current behavior and if it is not, it will perform the refill once, then ensure stamina.png is gone. If it is not, it will repeat the refill function (and thus increasing repetitions done so this will eat away an additional repetition. Hopefully this isn't a deal breaker).

Should work as intended and this behavior at the moment will work for every Region except "EN" which we can later change.

PR #146, opened 2018-12-24
Added function for JP Stamina Refill

TryBane:
Added a function that checks the region before entering refill stamina. In the instance it is true, stamina will be refilled as before, if not it will refill till the stamina image no longer exists on screen.

Needs to be checked in other regions to ensure whether it is needed for those as well. At the point of this original commit, it will only apply to EN as it is the only one I'm aware of that it would not be needed.

### DIFF
--- a/regular.lua
+++ b/regular.lua
@@ -62,9 +62,12 @@ function menu()
 
 	--Auto refill.
 	if StaminaRegion:exists(GeneralImagePath .. "stamina.png", 0) then
-		RefillStamina()
+		if GameRegion == "EN" then
+			RefillStamina()
+		else
+			RefillJP()
+		end
 	end
-
 	--Friend selection.
 	local hasSelectedSupport = support.selectSupport(Support_SelectionMode)
 	if hasSelectedSupport then
@@ -112,6 +115,13 @@ function RefillStamina()
 	else
 		scriptExit("AP ran out!")
 	end
+end
+
+function RefillJP()
+	-- Refill Stamina until stamina.png isn't on screen
+	repeat
+		RefillStamina()
+	until not StaminaRegion:exists(GeneralImagePath .. "stamina.png", 0)
 end
 
 function startQuest()
--- a/regular.lua
+++ b/regular.lua
@@ -218,6 +218,8 @@ function init()
 	setImmersiveMode(true)
 	Settings:setCompareDimension(true,1280)
 	Settings:setScriptDimension(true,2560)
+	
+	toast("Will only select servant/danger enemy as noble phantasm target, unless specified using Skill Command. Please check github for further detail.")
 
 	StoneUsed = 0
 	PSADialogue()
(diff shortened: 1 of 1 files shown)


potchy:
As to not increase code complexity, I think it's okay to use a single function for all servers.

Regarding the actual algorithm, shouldn't it check the AP screen again only once and throw an error if it is still there? I don't play the JP server, so correct me if I'm wrong, but if the AP screen is still there it means you don't have enough apples, so there's no reason to keep clicking, right?

potchy:
I watched the video and I still don't understand what's different for JP. lol

TryBane:
I watched the video too and I don't either, but he claims they behave differently when you are low on bronze apples. The change is pretty insignificant that I don't see it affecting performance at all, or maintainability. Any changes to the original Refill function will work as before, the only difference is for all other regions it will do 1 more check for the image for refilling and if it finds it, it will repeat the function again.

Other than that, no changes were made.

TryBane:
Yeah, that sounds about what the loop I implemented is supposed to do. And yeah, I guess making that change works a bit better regarding complexity. Though with refactoring I'd like to rename the function as "checkStamina" and move the loop itself into the function as well.

That being said, I'll commit those changes now and then merge as nothing else will really be affected.

29988122:
>Maybe it's like this?
Yes, that is the case. Probably.

I got overwhelmed by the immense amount of workload recently, so honestly I don't believe my memory that much. Also, it's impossible for me to use up my bronze apples to test - I've got like 300-ish thanks to the Christmas event 2018.
Current fix seems pretty good tho!

Also regarding NotJPserverForStaminaRefillExtraClick.
In the video you can see that the client entered the stage directly. IRRC, users still have to click on the stage to enter the stage on other servers. *This behavior change happened around the same time as the previous one.*

Is that still the case?

TryBane:
I'm not sure regarding CN or TW servers but NA had an update with the Anniversary that changed that behavior to be the same as JP, meaning it automatically enters the support selection screen after using an apple. Or at least I think it was Anniversary.

Might not be the case in CN or TW so I've kept that particular variable in there just to be sure.

29988122:
Thanks for the info.(good morning?) It's safe to delete them.
Pushed the commit.

Progress among servers: JP(chapter2) >>>>>>>>>>CN(chapter 1.5 ep2)>TW(CCC event now)>>EN(Dr. Roman!!!!!!!!!!!!!!!!!!!!!!)

TryBane:
Yeah, I just woke up though technically it is late afternoon for me. But I wasn't sure on their progress with Quality of Life updates. We are getting them about 1 year earlier than JP did so I was unsure whether CN and TW were getting them earlier or at the same pace as JP did. Either way, TW is the only one that might be needing it if they are a year behind JP so I doubt it will matter soon if that's the case.

TryBane:
Also, if CN and TW are beyond EN in terms of Quality of Life updates, then I should mention that EN also got the Fast Death animations. So there is no need for the click to skip death animations. This is assuming that people use the feature which you can toggle in the same menu as fast battle animations.

29988122:
>Fast Death animations
So they really separated major story updates from client usability updates...interesting.

That option only affects enemy death animation tho - if you're a new player, servants in your party dies easily. So I guess we could keep that because somebody asked for it, XD.

(Although the implementation is messy : (

Issue #109, opened 2018-12-01
 (Extra window appearing, script in wrong state issue)Permenant fix of script accidently clicked on extra status window


29988122:
Steps to reproduce:
1. Make Danger / Servant target switching active. (Default active if I recall correctly.)
2. Script enters Battle 3/3.
3. ONLY WHEN the danger target is at the right-most position of the battlefield, this bug happens.
4. The script will "switch" to the right-most target.
5. Since the game already has its default target on the right-most target, another click on it will pop out the enemy status window.
6. Stuck, yay.

Possible cause:
When you click on the target that you've already selected as the target, the enemy status window will pop out and f you up.

Workaround:
Just check if there's a cross icon (close window icon) on the certain position of the screen whenever you selected a target. Not recommended.

How to fix it permanently:
*Client language compatibility must be considered if possible*
There are 5 kinds of extra windows that will mess up our script routine.
1. Team wiped out window.
2. Master skill (master cloth skill) window.
3. Servant skill window (if the option's on/if the skill's on CD).
4. Enemy status window.
5. Servant status window.
Extra. DATA HAS BEEN UPDATED WINDOW (sometimes it happened during your battle, and, bang you've been sent back to the title screen. Thanks, Delightworks!)

I guess we'll have to figure out a way to handle them once and forever.
I have no idea how to write an Observer, or if it's possible in the Lua script engine. 
FYI: coroutine lib is not working http://ankulua-tw.boards.net/thread/691/bug-ankulua-coroutine

=
Side notes: I have no idea it's because I updated Nox to 6.2.5.3, or I updated Ankulua to 8.0.5.
The performance of recognition is VERY FAST now. You guys gotta give it a try.

TryBane:
You only open up that menu when you press and HOLD on them. I have never had an instance where just clicking on an already selected enemy produced the status menu for them. The lua scripts way of pressing on screen may produce a similar affect though.

This is just a bit of information to help if it does.

29988122:
@TryBane 

That's why we'll have to deal with it beforehand - sooner or later it will be applied to other clients.

TryBane:
Ah I see, you make a good point good sir!

PR #147, opened 2018-12-25
Try to fix GameRegion specific issue(target selection behavior in JP server)

29988122:
https://github.com/29988122/Fate-Grand-Order_Lua/issues/109
Refer to the latest comment to see the target selection behavior video.
The way to reproduce this bug was written in the first post of the issue.

It was long click before, but now it's only a single click. Damn annoying.

***

Apparently this needs your input.

1. I don't know how Lua transfer values inside<->outside functions, or modules. This GameRegion variable is from FGO_**_REGULAR.lua, so I have no idea it will work or not. 

2. More importantly, I have no idea if this is the right way to do abstraction - we'll have to achieve low coupling(software engineering) since this is the main purpose of the refactoring, right?

*(We have this, and a fair bit of region specific behavior being added frequently(filtering UI, button element position change, to name a few.), which at first implemented in JP server, but will probably be online in other servers within a year.)*

How do we handle these from now on?

### DIFF
--- a/modules/battle.lua
+++ b/modules/battle.lua
@@ -175,6 +175,10 @@ end
 
 chooseTarget = function(targetIndex)
 	click(TARGET_CLICK_ARRAY[targetIndex])
+	if GameRegion = "JP" then
+		click(TARGET_CLICK_ARRAY[targetIndex])
+		--Click again to close extra window. See https://github.com/29988122/Fate-Grand-Order_Lua/issues/109
+	end
 	onTargetChosen()
 end
 


TryBane:
An option for fixing this issue could be to use continueClick() option rather than a regular click. I just checked with EN version and it didn't have any adverse affects. You can just make it double click. You'll have to test it though since it is REALLY fast clicks, so be sure it isn't too fast (and if it is, try increase the amount of clicks by 1 or 2)

29988122:
Upon further testing, this fix will NOT work

Reason: my brain farted.
https://streamable.com/4yjoa

Still can't figure out a way to circumvent the issue.

TryBane:
As you said before, changes are being made in different region localizations at different times so this can become an unavoidable issue in the future. I think we just need to consider each situation as it occurs. Specifically, if it works in JP version, then it should work in future versions of 
EN as well.

So we program for the new implementations of JP first, then create workarounds for odd behavior in EN in comparison.

Also, you make a good point..... So we need to figure out when the current target is already targeted. Before, you said we could look for the cross mark for exiting a sub-menu, but said it is not recommended. Why is that not recommended?

29988122:
The size of extra window WILL CHANGE(with huge variation) as the length of the description text change. That means the position of crossmark varies a lot.
***
>different region localizations at different times
>work in future versions ofEN as well.

I agree with your view. However,

>Low coupling

The remaining question is just "how "we implement it. I can't quite figure it out. 

Add a few extra checks here and there, and the variable scope beyond the current module, doesn't seem to be a good idea - that's why my codebase's a mess before.

TryBane:
We may be able to get around that using the findAllNoExcept() function and increasing the potential region size for the window. I don't recall if it was particularly variable based on description size, but if that is the case this set of methods should fix it.

findAllNoExcept() returns a list of matches and doesn't throw an exception if not found so it won't stop the script in the middle and we can use it to click on the icon.

The use would look like this:

```lua
        icon = WindowCloseRegion:findAllNoExcept(GeneralImagePath .. "close_window.png")
        if icon then
                click(icon)
        end
```

To explain the above, icon is set to be equal to any matches given by findAllNoExcept(). If it returns nothing, then icon will be nil. In the if condition, if the condition is nil, it will default to false, meaning it will skip the click. If it is a match, it will default to true so it will enter the statement. And you can use a Match (which findAllNoExcept() returns) to use as a click location.

There may need to be some tweaking based on how it works.

potchy:
Can't you click again outside of the window's bounds? Most windows in the game can be closed this way.

TryBane:
That may be an option. Perhaps clicking on the master picture would work? I don't believe there is any behaviour in doing so that will adversely affect any past or current versions.

TryBane:
Yeah, that is entirely an option for it. Also, as my friend is in dire need of a version that fits an odd aspect ratio, I have come up with a possible solution to the issue and a way to make it work on any device. I'm not looking to step on any toes, so do you have any issue with me posting a solution before you get to it Potchy? The method I would use would probably be different than what you're working on since it wouldn't be using a different module really. If you don't want me to do that I can just simplify the solution so that my friend can use it temporarily until a permanent fix comes into play.

potchy:
Feel free to push a solution. Maybe it's more practical than my idea. Even if it isn't we can always improve or replace it later.

By the way, I'm going to travel on Wednesday, so I'll be away from my computer for a whole week. I may be able to reply to issues, but I won't be able to code anything for the time being.

29988122:
I fixed the issue in JP server partially, merge this discussion into https://github.com/29988122/Fate-Grand-Order_Lua/issues/118

PR #148, opened 2018-12-25
Resolution Scaling

TryBane:
I have implemented a makeshift Resolution Scaling formula that should allow for scaling to a different resolution for both having bars on the left/right and top/bottom.

This version ONLY has support for EN, as that is what I'm testing it with. Just need someone to test it in multiple different aspect ratios for me (since I STILL can't get nox to work with it).

I originally tried to get it to work with a module, but apparently I'm too dumb to figure out how a module works so oh well. If someone can separate it to the module that'd be cool. EN has the current correct formula, I believe the module I tried to make is not correct.

Also, I'd put this outside of EN Regular file but there is a Region at the beginning of the Regular files. I don't believe it is NEEDED there, so I'll probably move them all. But for all I know, TW needs it in a specific place.


TryBane:
Okay, so my friend has tried my version on two separate devices, one with 18.5:9 aspect ration, and another with 4:3 aspect ratio. The first had bars on the left/right, the second had bars on the top/bottom.

BOTH OF THESE WORKED WITHOUT A HITCH! :D

Also, added a commit to consider the change in JP where clicking on an already selected servant opens the status menu. I will add another commit so that it ONLY happens in JP in case people aren't using fastclick. If you guys want that condition removed that's fine.

TryBane:
Alright, I've made it possible to work in all the regions and adjusted the position of the toast. Also, I realized that it wouldn't allow people that had a normal aspect ratio play so I made a fix for that (I'm an idiot, don't worry it wasn't hard :D )

I've given it to a couple of people to test so hopefully everything works out well on their end! This will make 4 tests.

TryBane:
Made the Offsets global because I'm an idiot and need to be punished

29988122:
I guess since you've solved multiple people's resolution issue, you can merge this pull request and draft an release? (Dealing with that is also nightmare to early times Android UI developers - they have to layout 20+ before Android has a proper way to scale....just a hard job, so thank you.)

Thank you so so so much, taking care of elder + full-time job + side project must be very hard and tiresome...

TryBane:
Thanks man. Just wanted to also mention that I made an adjustment in here that may potentially fix the issue regarding the sub menu popping up. Might want to make an issue where we can discuss it should it not be solved. Feel free to try it to see if it works!

Also, the issue remains the same for those that have a Notch. Now I have a solution for those people, but I don't have an automated solution. Potchy might be able to come up with something but I don't even know if it is possible to detect when there is a Notch to begin solving this problem.

With that in mind, I do have an idea of a way to solve this just not an idea of how to code it. My suggestion is to either create an issue or put a section in the readme with some instructions on how to handle resolution if they have a Notch. And while we are at it, providing some instructions on getting the script to work on Android 8.0+ would be a nice touch either in the readme or issue or documentation. 

Regarding each issue, just so I can put it in 1 place, I will post each one of those here to be added later either in readme format, documentation, or an issue (whichever you feel is the more effective method of getting the information to the users)

Notch Resolution issue:

If you have a notch, the current method to solve this is to get the left hand side bar in pixel size and hardcode that value into xOffset towards the top of the Regular.lua file they use. You can derive this value by taking a screenshot with the game open, opening it in an image editing program (I use online image editor). Resize the image to a height of 720 (keep aspect ratio). Then measure the pixel count from the left to the edge of the actual FGO image. Take that value, double it, and use it as the hardcoded offset value and it should all work smoothly.

One thing to keep in mind with this band-aid solution is that rotating the phone 180° will beak the offset value so you will want to caution users to only use the device in that specific orientation if they don't want to break their script. Once we can figure out a way to automate the calculations and do them every time then we can not worry about orientation.

Now for Android 8.0+ users. They need Ankulua 8.1 or higher. From there, they can choose Media Projection instead of Daemon for their screenshot option and set their Accessibility options for clicking (make the down times a lower value) so that accidental holding of the skill buttons don't occur. Now they do not need to use USB debugging for this version but this comes with some issues.

The above mentioned issue where a skill click can last longer than intended resulting in a click and hold action rather than a singular click action HAS occured though that particular user said adjusting his click settings fixed it so it might not be an issue if people follow those instructions. Additionally, preferred support selection has an issue where swiping isn't working. I'm not sure if Android 8.0+ has an issue with Ankuluas manualTouch function that we use for that particular area but the current solution to this issue is reducing swipes of preferred selection method to 0 and increasing refresh times accordingly to account for not finding the servant they want quickly.

29988122:
Find out some time, revising readme.

Post this here for further ref:
```
8.0.0:

* new methods
    - addEditPassword()
* Support accessibility mode for touch method for Android versions 7.0 and above. Don't need either root or daemon.
    - Note: Some emulators (like BlueStacks and Nox) don't support Media Projection for screenshot. Therefore root or daemon is still needed.
    - Methods supported in Android versions 7.0 and above
        click(), swipe(), zoom(), longClick(PRML, seconds), doubleClick(), keyevent(3, 4 or 187), continueClick(),continueMultiTouch()
    - Methods supported in Android versions 8.0 and above
        dragDrop(), longClick(), manualTouch(), touchDown(), touchMove(), touchUp()
    - Methods not supported in accessibility mode
        type(), keyevent(except 3, 4 or 187), killApp()
```

ryuga93:
> If you have a notch, the current method to solve this is to get the left hand side bar in pixel size and hardcode that value into xOffset towards the top of the Regular.lua file they use. You can derive this value by taking a screenshot with the game open, opening it in an image editing program (I use online image editor). Resize the image to a height of 720 (keep aspect ratio). Then measure the pixel count from the left to the edge of the actual FGO image. Take that value, double it, and use it as the hardcoded offset value and it should all work smoothly.

I think it is safe to assume that the left bar is in black, and for top and bottom bars will have the same offset to position the UI in the center of the screen. So, how about trying to detect the black color for the bar, and calculate the xOffset manually using a while loop:
```
if yDifferential > xDifferential then
        snapshotColor()
        local x, r ,g, b = 0, 0, 0, 0
        while (r < 3 and g < 3 and b < 3)  do
            r, g, b = getColor(Location(x,100))
            x = x+1
        end

        usePreviousSnap(false)
        xOffset  = x / yDifferential 
end
```

TryBane:
screenshot_20181226-072257
Example while I was trying to help him debug.

ryuga93:
I like your idea. We can use yDifferential and xDifferential to check which axis is off the 16:9 ratio, then setDimension(), and use the proper axis to limit the search region along with getAppUsableScreenSize()(:getX() or getY()) / 2. 

Then using the distance from the image the edge in 16:9 ratio, we can get the distance from the image to the edge of UI, minus the bar size and calculate the offset.

I would suggest to save the offset to a file so that the script could directly read the values in the subsequent runs, and make offset calculation to run once only when the file not exist or the values are nil

ryuga93:
I think that depends on how much time the calculation needs, if it is unnoticeable, I'm fine to allow user to rotate, else just notify them to rotate the phone such that the home button is on the right.

TryBane:
If it is unnoticeable i completely agree. We can even wrap the image checking functions within a wrapper function that first tests for what side the screen is rotated before proceeding. For instance, should they constantly rotate between skill activations it will check between each one of the activations for the battle screen location using each potential offset and then change that offset to fit till the next image check.

ryuga93:
That's a bit overkill...Though I have a weird feeling that someone will make this request in future. I'd prefer to ask them to decide the rotation before running the script, be it 0° or 180° and lock to that.

29988122:
>@ryuga93  I have a weird feeling that someone will make this request in future

Indeed..Since this project's for end-user, not for developers. xD

29988122:
Ahh fuck there are so many 18:9 phones 

I feel old using htc 10 without Notch

29988122:
Yeah, could be in the troubleshooting part, or even Android 8.1 part - since now 16:9 resolution phone is *minority*, so we should added that and Notch FAQ.

Happy new year!

ryuga93:
よいお年を!

Waiting to watch the Fate Project Special

PR #169, opened 2019-02-12
Reversed autoChooseTarget order

potchy:
Just checking if everyone agrees before merging.


potchy:
I think we should start with getting screenshots. :c

29988122:
Remember continueClick(game.RESULT_NEXT_CLICK,45) ?

It will often skip the 1st step of CE reward screen, thus it will be hard to determine where the screen lies in the whole result flow.

***

https://github.com/29988122/Fate-Grand-Order_Lua/wiki/In-Game-Result-Screen-Flow

The problem is: where does it appear, not what the screen looks like. (they looked all the same.)


For example, there's CE gain from "Event Points", and they don't get into your giftbox, instead they pop out directly after battle.

I still don't know where will that appear, since our current detect mechanism can handle servant CE gain properly, but not this CE gain behavior.

potchy:
Well, we can always place those in the main loop.
Now that it uses `ankuluaUtils.UseSameSnapIn`, I think it can handle many conditionals without affecting performance too much.

PR #188, opened 2019-07-08
Add updating click for select first method on select support screen.

aqer:
For current JP server, if user filter out 5* MLB event CE and use select first method,
it may not show any support and stuck at the support screen.

### DIFF
--- a/modules/support.lua
+++ b/modules/support.lua
@@ -84,6 +84,16 @@ end
 
 selectFirst = function()
 	click(game.SUPPORT_FIRST_SUPPORT_CLICK)
+	while game.SUPPORT_SCREEN_REGION:exists(GeneralImagePath .. "support_screen.png")
+	do
+		wait(10)
+		click(game.SUPPORT_UPDATE_CLICK)
+		wait(1)
+		click(game.SUPPORT_UPDATE_YES_CLICK)
+		wait(3)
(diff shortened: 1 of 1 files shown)


29988122:
selectFirst should be a fall-back way of all the friend selection behaviors, and should be executed efficiently because of this reason.

The wait(10) in the beginning of the script does not seem like a good idea.

Do you really need to save every bit of stamina? 
I myself won't choose MLB because of the problem you've encountered.

***

In the newer version of Ankulua, an observer-like function is implemented.
This can further be used to improve the efficiency of FGO-LUA (no more stuck and exception can be handled more elegantly), but I'm afraid that all 3 of the collaborators don't have much free time. I'm sorry : (

aqer:
No matter selectFirst is fall-back from other method or not, I only change the behavior when there is no support on the screen in selectFirst. If there are some supports on the screen, the behavior will be the same with current trunk.

The wait(10) is used for the interval of refreshing support list. It seems to me that wait(10) for refreshing support list is better than stuck at the screen and do nothing in current behavior.

29988122:
Sorry for being stupid - I did not write this module, it was @potchy 's work. 
I recheck the code, and your modification is indeed useful & no halting as you've described.

Thank you very much, @aqer !

PR #194, opened 2019-07-13
double check support_screen.png to satisify slower environment.

### LINKED ISSUE #192: Scrip stuck after selecting Support, on both servers JP/NA
Since the 7/8 Commit, "Add updating click for select first method on select support screen.", when farming for QP in both servers (NA and JP), it goes correctly to choose the 1st support option, but after that it gets stuck in the screen when you click to "go into battle".

It does the movement if as it was updating the support list, this results in swapping 4th and 6th servant, then it press "ok", but after that nothing happens, you have to manually click "go into battle", then it continues normally until again it has to select support and start another battle.

Tested on both JP/NA, both in Nox and Android phone, the issue happens in both root (phone) and daemon (nox and phone).

Using the 7/5 commit works without issues.

Nox is using 5.1.1, Android phone is using 7.1.2

aqer:
for slower machines, double check support_screen.png 
Fixes #192

### DIFF
--- a/modules/support.lua
+++ b/modules/support.lua
@@ -84,15 +84,18 @@ end
 
 selectFirst = function()
 	click(game.SUPPORT_FIRST_SUPPORT_CLICK)
-	while game.SUPPORT_SCREEN_REGION:exists(GeneralImagePath .. "support_screen.png")
-	do
-		wait(10)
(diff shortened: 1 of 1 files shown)


29988122:
Thank you for your help maintaining modules!

PR #197, opened 2019-07-17
added Skill_Confirmation == 1 check to OrderChange()

forgehe:
fixed bug with script that fails to do an order change without clicking OK on Skill Confirmation if user has that option enabled.


29988122:
Thank you for the bug fix!

Issue #212, opened 2019-08-17
(Request) Select friend support by friend name and class besides servants/CEs 


muvluvalt:
It would be nice to have these 2 features, especially when one is going through the Main Quest and the classes differ for each node (meaning that the Servant/CE selected would vary immensely). The current option for this is just to set support selection to manual, but that kind of defeats the whole purpose of automation. Setting support selection to first is also not a viable option, because that would then select the NPC support instead of friend support.

Therefore, automatically selecting support by friend name, and by class, would help to mitigate this issue by selecting the desired support without having to take multiple Servant and CE screenshots.

In other words:

- In an archer node, user will be brought to the Lancer class section of the support list.

- In a mixed node (eg. Saber,Lancer), user will be brought to some random class section with class advantage over one class in that node (eg. Archer) of the support list.

In the latter case, selecting the Berserker class may be the better option since Archers are also weak to Lancers. The script doesn't switch classes (i.e from Archer to Berserker), which is one of the two features I'm requesting. 

P.S I'm not sure if this feature has already been implemented, so if it's already implemented, I will close this issue.

TryBane:
I believe I wouldn't have an issue setting the script up to select a support by friend name, however you would still have to create a screenshot of that friends name. Otherwise we would literally need to create a library of images for every single letter of the alphabet AND any other language to use when you are selecting the support.

The reason we would need that, is because there is no limit to the characters people can use when creating their names. You are only limited in the number of characters.

As for the selection of the class, I believe that one might be a bit further than we would like to go. Because small UI differences between versions regarding their position, size, etc can cause incompatibility between versions. Example, say they introduce a new tab for supports. But now this tab needs to be squeezed in with the current line of them. This will either expand the size of the region needed to cover, which would be the ideal scenario, or the more likely instance would be them shrinking the size of all the other tab images to fit the new one. And this would require a replacement of all the current images. Not only that, but it would result in needing multiple copies of them to cover all versions, both the ones that have the new tab and the ones that dont. 

Thus is my reasoning for why I believe it would be very burdensome to implement and then maintain.

In conclusion, I could very easily implement a system that can select a servant purely based on name, but this would still require screenshots to be taken and adjusted by the user, such as yourself. But I may implement this anyways, because it sounds like something I would likely use as well. But the others would be difficult to maintain and excessively inflate the size the script takes up on anyone's device.

muvluvalt:
[...]

PR #216, opened 2019-08-23
Friend Specification for Support

TryBane:
As requested by someone, I put together a new feature for Support selection. With this, people can make a picture of a friends name on their support list and use that to find said friend instead of a specific support. They will need to ensure they are in the right class selection though, to make sure they are granted the right support they expect.

With this, the Support selection process doesn't change for the most part. There is a new option in the FGO_XX_REGULAR.lua files for listing the names of the friend images they make, just like how you list the Servants/CEs you are looking for. You will use "friend" word instead of "preferred", "first", or "manual".

I used the same process used for the preferred option, just taking out some unneeded case handling since there are no instances of needing to match CEs or anything.with the names. I have tested to make sure it works as intended in NA server, should be fine in the others as well. I have to adjust the readme still, but that shouldn't be difficult. The original request is in #212 along with an example of what screenshotting the name looks like. I would like either a review of the changes or testing of them to ensure nothing seems out of place.

After this gets approved, I will begin work on a couple other features in preparation for the Lotto events coming in all the servers. The two I'm going to work on for a friend are selecting certain number of cards before NP and choosing targets between skills/turns. A friend needs these for his planned team for Nerofest on NA and I feel it would be useful for everyone if they have the option.

### DIFF
--- a/modules/support.lua
+++ b/modules/support.lua
@@ -79,6 +79,9 @@ selectSupport = function(selectionMode)
 
 		elseif selectionMode == "manual" then
 			selectManual()
+		
+		elseif selectionMode == "friend" then
+			return selectFriend()
 
 		elseif selectionMode == "preferred" then
 			local searchMethod = decideSearchMethod()
@@ -112,6 +115,14 @@ end
 
 selectManual = function()
 	scriptExit("Support selection mode set to \"manual\".")
+end
+
+selectFriend = function()
+	if #FriendNameArray > 0 then
+		return selectPreferred(searchMethod.byFriendName)
+	end
+	
+	scriptExit("When using \"friend\" support selection mode, specify at least one friend name.")
 end
 
 selectPreferred = function(searchMethod)
@@ -189,14 +200,10 @@ searchVisible = function(searchMethod)
 end
 
 decideSearchMethod = function()
-	local hasFriendNames = #FriendNameArray > 0
 	local hasServants = #PreferredServantArray > 0
 	local hasCraftEssences = #PreferredCraftEssenceTable > 0
 
-	if hasFriendNames then
-		return searchMethod.byFriendName
-
-	elseif hasServants and hasCraftEssences then
+	if hasServants and hasCraftEssences then
 		return searchMethod.byServantAndCraftEssence
 
 	elseif hasServants then
@@ -206,7 +213,7 @@ decideSearchMethod = function()
 		return searchMethod.byCraftEssence
 
 	else
-		scriptExit("When using \"preferred\" support selection mode, specify at least one friend's name, a Servant or a Craft Essence.")
+		scriptExit("When using \"preferred\" support selection mode, specify at least one Servant or Craft Essence.")
 	end
 end
 
--- a/modules/support.lua
+++ b/modules/support.lua
@@ -24,6 +24,7 @@ local scrollList
 local searchVisible
 local decideSearchMethod
 local searchMethod
+local findFriendName
 local findServants
 local findCraftEssence
 local findSupportBounds
@@ -44,6 +45,12 @@ init = function()
 
 		return values
 	end
+	
+	-- friend names
+	for _, friend in ipairs(split(Support_FriendNames)) do
+		friend = stringUtils.Trim(friend)
+		table.insert(FriendNameArray, friend)
+	end
 
 	-- servants
 	for _, servant in ipairs(split(Support_PreferredServants)) do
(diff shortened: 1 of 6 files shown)


potchy:
I've tested it and it works great!
I only disagree with the need to set `Support_FallbackTo` to `friend` instead of keeping it on `preferred`. I think it will be easier to use if we just ignore the other options when `Support_FriendNames` is filled. It's one less option to worry about. I'm not sure I was clear enough, so please just check the changes I made to `decideSearchMethod`. lol

```lua
decideSearchMethod = function()
	local hasFriendNames = #FriendNameArray > 0
	local hasServants = #PreferredServantArray > 0
	local hasCraftEssences = #PreferredCraftEssenceTable > 0

	if hasFriendNames then
		return searchMethod.byFriendName

	elseif hasServants and hasCraftEssences then
		return searchMethod.byServantAndCraftEssence

	elseif hasServants then
		return searchMethod.byServant

	elseif hasCraftEssences then
		return searchMethod.byCraftEssence

	else
		scriptExit("blablabla")
	end
end
```

<sup><sup><sup>I also took the change to merge some duplicated code</sup></sup></sup>

modules/support.lua
```diff
@@ -44,6 +45,12 @@ init = function()
 
 		return values
 	end
+
```
potchy: moved this up, but didn't change anything

modules/support.lua
```diff
@@ -63,12 +70,6 @@ init = function()
 
 		table.insert(PreferredCraftEssenceTable, craftEssenceEntry)
 	end
-	
-	-- friends
-	for _, friend in ipairs(split(Support_FriendNames)) do
-		friend = stringUtils.Trim(friend)
-		table.insert(FriendNameArray, friend)
-	end
```
potchy: moved this up, but didn't change anything

modules/support.lua
```diff
@@ -319,37 +300,6 @@ isLimitBroken = function(craftEssence)
...
-	local friends = {}
-
-	for _, friendName in ipairs(FriendNameArray) do
-		for _, theFriend in ipairs(regionFindAllNoFindException(game.SUPPORT_FRIENDS_REGION, Pattern(SupportImagePath .. friendName))) do
-			table.insert(friends, theFriend)
-		end
-	end
-
-	return friends[1]
-end
```
potchy: moved this up, but didn't change anything

modules/support.lua
```diff
@@ -319,37 +300,6 @@ isLimitBroken = function(craftEssence)
...
-		if support == nil then
-			return {"not-found"} -- nope, not found this time. keep scrolling
-		end
-
-		return {"ok", support}
-	end
-
-	-- see https://www.lua.org/pil/5.1.html for details on "unpack"
-	return unpack(ankuluaUtils.UseSameSnapIn(performSearch))
-end
```
potchy: removed

modules/support.lua
```diff
@@ -151,42 +150,6 @@ selectPreferred = function(searchMethod)
...
-			numberOfUpdates = numberOfUpdates + 1
-			numberOfSwipes = 0
-			
-		else
-			click(game.SUPPORT_LIST_TOP_CLICK)
-			return selectSupport(Support_FallbackTo)
-		end
-	end
-end
-
```
potchy: removed

modules/support.lua
potchy: moved this up, but didn't change anything

TryBane:
If I'm understanding correctly, you mean that we just ignore fallbackto altogether? I only had an issue with that idea because people not logging in often can cause issues with the amount of time it takes to find them. I felt that providing them the option to make it an infinite loop with friend in both options would allow them to decide that it is what they specifically want or to fallback to finding a specific servant/ce if it is taking too long. Though I don't necessarily think your idea is a bad one.

But thanks! I also have targeting done as well! I have had multiple people test it on NA and JP and it works like a charm! After this is done getting reviewed, I'll be pushing that one as well! It'll be great to have you test that as well.

Also, we need to update the readme after these things are added. The script can be started in the support list and results screen but those options aren't listed at the moment. Then there is this new option once it is approved and the other new features that'll require the Autoskill portion updated as well.

potchy:
Hmm, you're right. I completely missed your point.
Reverting~ I think this can be merged now.

TryBane:
It looked a lot nicer though! But thanks, I'll go ahead and merge and set the other up for push.

PR #220, opened 2019-08-25
Targetable

TryBane:
I have added functionality to target specific enemies using autoskill. To do so, simply include 't' and a number to indicate which enemy you want to target. 1, 2, or 3. The format for enemy targets based on screen are:

t1    t2    t3

So an autoskill such as a2t3ht1i would use 1st skill on 2nd servant, target the far right enemy, use 8th skill, target far left enemy, use 9th skill.

This has already been tested by a few of my friends including myself and has been found to work on both NA and JP servers. Anyone who wishes to give it more tests, feel free. And if you could double check my logic, that would be great.

Another note, the auto target is still active, and won't be interfered with using this, but I believe we should make it set to 0 by default in order to avoid confusion or possible mishaps with the autoskill. Essentially, the script will assume you are using autoskill to target first, then if you decide you don't wish to, you can toggle auto target on for simpler autoskill command lists.

### DIFF
--- a/modules/autoskill.lua
+++ b/modules/autoskill.lua
@@ -103,6 +103,17 @@ local function SelectSubMember(location)
 	end
 end
 
+local function SelectTarget()
+	ChangeArray(ENEMY_TARGET_ARRAY)
+end
+
+local function SelectEnemyTarget(location)
+	click(location)
+	click(game.BATTLE_EXTRAINFO_WINDOW_CLOSE_CLICK) -- Exit any extra menu
+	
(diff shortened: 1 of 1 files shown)


potchy:
You nailed this one!

potchy:
> Another note, the auto target is still active, and won't be interfered with using this, but I believe we should make it set to 0 by default in order to avoid confusion or possible mishaps with the autoskill.

I think it's better to leave it as is, because that's the behavior people are already used to, myself included.

PR #221, opened 2019-08-25
Pre load np

TryBane:
This one I had a tough time naming. The functionality of this is allowing people to use cards before NP with Autoskill. The command is n1 and n2, for 1 and 2 cards respectively. The changes include reforming the clickCommandCards function to include 2 parameters, one for the number of cards clicked already and another for number of cards that need to be clicked. The second change is in battle.lua, where I added a new function to inform the module of how many cards have been clicked, and updated to accommodate the change to clickCommandCards. And the third change is in autoskill.lua itself where I added the functionality.

clickConmandCards now checks if the number of previously clicked cards has been passed by using an index counter, then when it is passed that point it will click again. It will also automatically return once the index counter reaches the clicked cards amount that is expected.

An example of this in use would be "a2t3ct1n24". Essentially since using this functionality opens the attack button, it needs to be done right before NPs are clicked or you will miss the number of skills before and the script will freeze and be stuck.

You can only select 1 or 2 though, since the idea is that it preloads the NP with cards (important for certain comps that need an enemies HP weaker for more Overkill for NP gain from NP) meaning an NP card should be clicked.

One final note as well, this will be clicking in the order you specify in the config file. So if you want ABQ order, it will follow that same priority.

EDIT: I have also tested this myself on NA, and have had others test it on both NA and JP as well and so far it seems to work as it should on both servers.


vit2657:
Testing this branch now, everything seems to work, but getting toast about unable to determine a card type at some position from time to time even though it works.

TryBane:
Yes, the failed to determine card type is normal, as it is trying to figure out the cards that are selected. But due to being selected, they aren't being registered. This isn't an issue, it is normal.

potchy:
Please fix the conflicts before we merge it :c

TheRealPlagiariser:
Sorry to bump this, but why has this not yet been integrated in the main release? Working fine on Pixel 2

TryBane:
Potchy mentioned conflicts... But it says there are no conflicts with the Base branch? Not sure what the problem is....

reconman:
@TryBane I noticed your merge request and when you look into the "Files changed" tab, it says that all lines were changed because you changed the end of the lines from \r\n (Windows style) to just \n (Linux style).

Can you fix that please?

reconman:
Works perfect for me, there are just small whitespace formatting issues and I don't like that users get that "Failed to determine Card type" message because it confuses users.

Is there any way to avoid that message by adding some if conditions or something like that? I don't have much experience with Ankulua or FGOLua, so I can't figure it out myself.

Also I wouldn't call it "preload NP" but rather something like "attack before NP" because preloading the NP would mean that the NP is the first thing that happens or that it's prepared before you attack.

TryBane:
I am closing this Pull Request because I have overhauled the feature and am going to Repull it separately.

Issue #206, opened 2019-08-08
SUPPORT_LIST_ITEM_REGION_ARRAY?


RalphixJP:
I'm getting this error every hour when it's trying to select my preferred support servant from my friends list. 

![Screenshot_2019-08-07-22-36-08](https://user-images.githubusercontent.com/15004197/62671272-2fb88200-b964-11e9-8660-fc7084137c99.jpg)

muvluvalt:
I'm also experiencing this problem in the JP server, but only when I'm selecting Waver with Teatime CE (default image inside IMAGE_SUPPORT). I don't have any trouble with any servant + Mona Lisa Lily CE (custom screenshot by me). Note that the image file name of teatime is "teatime.png" without the asterisk in front.

```
Support_PreferredServants = "waver4.png, waver3.png, waver2.png, waver1.png"
Support_PreferredCEs = "teatime.png"
```

```
Support_PreferredServants = ""
Support_PreferredCEs = "*mona_lisa_lily.png"
```

TheIanator:
The script needs to scroll in precise intervals to work. It recently stopped working for me as well (on JP, haven't tested NA). Don't think it's a problem with the script as there's been no changes to how it scrolls the support screen and scripts that were working two weeks ago fail today.

So it's probably either a change to the game or a change to AnkuLua. Maybe downgrading AnkuLua might fix the problem like in issue #205.

muvluvalt:
The issue first occurred to me when I was at version 8.5.0. That was when I upgraded to version 8.5.2 but the problem persisted. Downgrading to any version below 8.5.0 would make the script completely unusable, which was why I didn't do it.

https://github.com/29988122/Fate-Grand-Order_Lua/issues/187
https://github.com/29988122/Fate-Grand-Order_Lua/issues/185

RalphixJP:
> Actually, never mind. It was because I didn't put the asterisk in front of the file name. Can't believe that I made such a dumb mistake.
> 
> `Support_PreferredCEs = "*teatime.png"`
> 
> Edit: No wait, problem still persists even after I added the asterisk. Downgrading the version also doesn't work.

Adding an * only makes it so that the CE the support has is MLB.

TryBane:
This may very likely be due to a slight UI change with the anniversary update. I will need to further investigate the issue before I can say definitively if I have a fix.

TryBane:
[...]

Issue #106, opened 2018-11-28
Issue selecting supports


raspberrytealeaf:
Hiya,

I'm using the waver face cards and holy night supper included from your files. It works the majority of the time, but sometimes I will get an error:

![image](https://user-images.githubusercontent.com/7444536/49131684-559a6c80-f296-11e8-9055-2d2da89287dc.png)
![image](https://user-images.githubusercontent.com/7444536/49131694-5a5f2080-f296-11e8-888d-5605411259b6.png)

I have no idea what triggers this, but I think it's related to scrolling down / swiping on the friends list. When there's a Waver w/Holy Night Supper at the 1st or 2nd slots, it seemingly has no issues selecting him.

Any suggestions on how to fix this?

ryuga93:
The error happens when the scrolling reached the end of the support list, causing the boundary to fall out of expected range.

You can reduce the value of `Support_SwipesPerUpdate`, maybe to 4 to fix it.

@potchy Maybe instead of throwing the error, how about catching it and toast it as notification and proceed to update the support list?

raspberrytealeaf:
Ah, I see. I'll reduce the swaps. Thanks.

potchy:
Just to confirm: the error appeared when the script found the right match in the 3rd support slot? I mean 3rd visible on screen, not total.

ryuga93:
![image](https://user-images.githubusercontent.com/6499824/49153149-e2edc780-f34f-11e8-896b-977f9f3de207.png)
It throws error when the second visible servant is matched regardless the CE. For the image above, I'm trying to get Ishtar with the 5 Start Gacha CE.

raspberrytealeaf:
@potchy It seems to occur after it swipes down a few times and needs to refresh cause it doesn't find the servant + CE you want. I think what @ryuga93 said is on point. My screen res is whatever the default NOX resolution is.

TryBane:
@raspberrytealeaf While working on a fix, it might be better to limit the number of swipes in the settings in the meantime to keep it from throwing that error.

potchy:
I just pushed a fix for this issue.
Please help check if it has indeed been solved before I close it. 😄

raspberrytealeaf:
I've been running the updated script for a while w/more siwpes and so far I haven't encountered the same error. Thanks!

PR #222, opened 2019-08-26
Why is this here?

TryBane:
Looking into the issue regarding Support scrolling, and I'm not sure why this is here. In any case where the support items aren't found, it should return nil, and in the case where it returns nil, it'll scroll further. And when it hits the bottom, it still shouldn't crash.


potchy:
It's in there to indicate that a servant/CE was found out of the expected bounds, i.e. either `scrollList()` is miscalibrated (#206) or `SUPPORT_LIST_ITEM_REGION_ARRAY` needs some adjustments or the coordinates got misaligned because of the notch calculations and stuff. It's complicated. See:

#106 
https://raw.githubusercontent.com/29988122/Fate-Grand-Order_Lua/master/docs/support_list_item_regions_top.png
https://raw.githubusercontent.com/29988122/Fate-Grand-Order_Lua/master/docs/support_list_item_regions_bottom.png

I'm fine with "fixing" the error, but you should replace it for a toast, instead of removing it altogether. This way, it will still indicate that there's something wrong instead of failing silently.

TryBane:
But it isn't really failing due to any issue with the script, it is failing due to something akin to lag that is probably the result of Media Projection (as I haven't seen this issue occur much at all from android 7.0 and lower)

A toast would probably be fine, but when the image is present but not within the bounds because of a notch, it does fail with that error, it sits there and does nothing. The only time this error occurs is if the Servant is found at the bottom but due to the space the CE can't be checked. And then what results is this error, even though after scrolling it would be able to see if the CE is there afterwards. I have yet to see another instance of this case coming up in another scenario.

potchy:
These 3rd and 4th regions should account for the bottom of the list.
If it's not working then there is definitely something wrong.

https://raw.githubusercontent.com/29988122/Fate-Grand-Order_Lua/master/docs/support_list_item_regions_bottom.png

potchy:
If it's lagging that much, try to tweak the wait time here:
https://github.com/29988122/Fate-Grand-Order_Lua/blob/231971110b2f5de334b92ddcd7b4f6d54068a23d/modules/support.lua#L170

TryBane:
The issue the lag causes isn't a recognition issue but rather an issue in how far it ends up scrolling. The lag disrupts the scroll distance causing the menu to not scroll enough and thus disaligns the servants found from the region array provided. I believe an alternative to the region array would be to create the bounds dynamically after it recognizes the servant. I will look into this, but I do realize why this error message exists.

I believe I have come up with a solution just now. I will see about implementing it in a couple of hours (if you'll still be on to double check my work)

potchy:
If you disregard the region array and calculate the distance to the CE dynamically while assuming that `scrollList()` is inconsistent, which is what I tried to do while writing this module for the first time, you'll also run into the problem of some servants/CEs being skipped and you'll need to scroll up/down accordingly, otherwise the script would never select those.

TryBane:
Actually, I was going to dynamically create the regions myself. I'd maintain the same exact usage, but by creating the regions dynamically after the scroll finishes each time, it won't run into an issue where they don't align with the servant options, thus eliminating the need for the exit message.

But, I'll run down what I plan to do. I am going to make a small image that is searched for that will determine the location a support (I can use the Quick card icon on the right side) and the from the location of that image I can extrapolate the entire segment, thus creating the region. Then I can make a duplicate above or below the one already made based on whether it would go out of bounds or not.

Then I can use these two regions (only 2 will ever be able to be made) to create the array used in the current search formula.

As a result, we can remove the error message and still be able to enjoy the search without any worries of lag causing an inconsistent scrolling issue or hitting the bottom of the friends list.

Due to the nature of the scrolling as it is right now, it should never scroll in such a way that a servant or CE is missed.

TryBane:
@potchy Okay, I am pretty confident that this should work as intended. Give it a try and see if you are satisfied with this as a replacement.

MatheusOPP:
I'm a beginner to GitHub so I'm not sure if this is supposed to be correct procedure, but I tried replicating the exact changes in the support.lua file to reflect the commits above, and copying the new image to its appropriate folder to try out this solution, since with the files that are currently in the master branch, I end up getting an error after the script runs for some time (around one hour or so) in the support screen that's the same as #229.

However, doing exactly that is giving me an error where after two swipes or so in the support screen, the script returns another error, every single time. I attached screenshots of the error below, if you want to check it.
error1
error2
error3

TryBane:
You should have replaced the files as mistakes are bound to be made. Can you upload the file you have with the changes so I can take a look?

MatheusOPP:
Sure. Had to zip it since GitHub doesn't let me attach it directly here, but here it is. Also put my "FGO_EN_REGULAR.lua" there just in case.
[support.zip](https://github.com/29988122/Fate-Grand-Order_Lua/files/3563303/support.zip)

TryBane:
Okay, I see I see. I didn't consider that returning nil would break the search algorithm. Will have to fix that. I know of a way but can't do it till I get home. Give me about 30 minutes to solve this and if you could try again that would be awesome.

TryBane:
@MatheusOPP Okay, change made. That should work as long as I didn't make a spelling mistake (which shouldn't be the case since I did copy paste....)

potchy:
I liked your idea and I believe it will work great (will test right away), but now **I'm** the one who doesn't understand why this conditional exists.
https://github.com/29988122/Fate-Grand-Order_Lua/blob/383483c8ab2df53659ea2379e6d35b6edcce2364/modules/support.lua#L295-L296

Also, @MatheusOPP problem indicates that there's some scenario we didn't think of yet.
Your last commit (383483c) doesn't fix the problem. It's just swallowing the error and failing silently.
I'm not sure if it's related, but maybe you could this as an anchor instead, as it's closer to the top of the list item.

image

TryBane:
I setup that conditional so that it doesn't return a support that has the CE hidden on the bottom. However, I realize that that won't work as well as needed for when people are just looking for the Servant so it probably shouldn't exist.

As for my commit to solve @MatheusOPP's problem, it occurred when searching for both a Servant and CE, but then the Servant was found on the bottom when the CE is hidden. When this occurs, it'll try to get the bounds and return it, however due to that conditional, it'll kick it out and return nil instead. My commit allows it to still return a bounds, just not one with the Servant or CE, and is on the top, which will nullify any "potential" false positive match with CEs due to the conditional that the CE be below the Servant found.

If we remove the conditional, then we probably don't need to have the latest commit either. Glad you like it too.

MatheusOPP:
No sure if my input is still needed at this point but, I let it run for some hours now and it seems to be working perfectly. I'll let you know if anything strange happens later.

potchy:
I think it's okay to merge this whenever you're ready. Just please read my previous comment again as I edited it multiple times and I'm not sure you got everything. lol
Nevertheless, I may revisit the search algorithm when I do get some free time. Work has been crazy this month.

TryBane:
I see, I understand the last paragraph (which wasn't there when I commented I believe) but I don't believe it is necessary. It would probably work of course, but by removing the conditional it shouldn't be an issue (and should still allow us to remove the default return bounds). I'll go ahead and do that real quick and then merge.

TryBane:
Also took your advice and made it toast in case it hits the default return, as something went wrong in that case.

PR #285, opened 2019-11-17
Support selection fix (mainly for JP)

vit2657:
Currently support area recognition is using small quick card to find an area. This is a problem on JP however, due to some times support servants can have a command code in their cards, which leads to a default region being selected (and in my personal case it leads to a wrong support being selected).
Here I have added another image and updated regions. Tested on JP.


vit2657:
So far tested and works perfectly on JP.

However, this is not a perfect solution and I encourage you not to merge it if want to accept this.
1. I've expanded the region width to 2356 which is too big to find that exact support image. Ideally it should be much smaller, but I lack measuring tools to make it perfect. Any assistance on that would be awesome.
Also the region numbers itself should be moved to this file after the fix (I think): https://github.com/29988122/Fate-Grand-Order_Lua/blob/eea7a51a0b6c39c24610d2cb6df032b645f071b8/modules/game.lua
2. Can't test on NA any time soon, if some one can, I will appreciate this.

TryBane:
Sorry about the delay, but I have actually already fixed this with one of the branches. I was waiting for someone to test it though before I fully merged it. I must've forgotten to pull it.

https://github.com/29988122/Fate-Grand-Order_Lua/tree/Support-C

TryBane:
@vit2657 I have a new pull request pending that fixes the Support selection problem for JP. This solution isn't complete though, as it won't work on TW and CN until I have the appropriate images for them. But you can feel free to take a look and give it a try for JP. I'll provide the link to it here #335 so it is easier to navigate to.

TryBane:
With the most recent merge, this pull request should now be redundant

PR #293, opened 2019-12-07
Auto Gift Box

TheIanator:
To help those trying to maximize storage capacity of their gift box. Scrolls through the list automatically checking the boxes of all silver XP and any gold XP in a stack of fewer than 3.

Can change the variable "goldThreshold" to a different value if desired.


reconman:
Doesn't work on my device with 2280x1080 resolution. Even when I adjust it, 1s are detected as 4. When I added a `Settings:set("MinSimilarity", 0.9)`, 3s were also detected as < 3.

I can't figure out how to fix the OCR problems.

TheIanator:
This version is currently set for JP, I should do an update and add a setting for which server and maybe a couple of other things.

If you want to get it working in the mean time replacing "countRegion = Region(355, gift:getY() - 50, 100, 30)" with "countRegion = Region(420, gift:getY() - 50, 100, 30)" should get it working.

Inkaflare:
This is something I was specifically looking for with the excess of Embers I'm getting, so thanks for that! 
I ran into some trouble trying it out. I downloaded your entire commit to get the needed image files and modifications of the other scripts and placed them on my phone, overwriting the existing files I had, except for FGO_EN_REGULAR.lua since I have been modifying it for my own use, but it shouldnt interfere with your new script. Yet I get the following error message when running it:

From my limited understanding, it seems like the checkmarks aren't being recognized in my game. Is this related to my device's resolution or did I make a mistake somewhere? 
Thanks for the help!

reconman:
@Inkaflare I will open some boxes later today and try if I can get a version that runs on NA on any resolution. It won't be pretty code, but it will work.

I can upload the modified file to my fork, then you can use it. I'll notify you when I have a working version.

Inkaflare:
Thanks for the quick reply. So I was right to assume this is likely a problem related to my resolution?
I'm not fussed about pretty code, I'm just tired of the slog that is manually claiming embers. Looking forward to it!

TheIanator:
I copied some code from _auto_gift_exchange.lua that I think was added to deal with resolution issues and switched the variable for NA. It still works for me, hopefully fixes it for odd resolutions.

[Fixed_Auto_Giftbox.zip](https://github.com/29988122/Fate-Grand-Order_Lua/files/4083114/Fixed_Auto_Giftbox.zip)

Just replace your _auto_gift_box.lua and hopefully it will work. I also included _auto_multi.lua which is just a pop up box that asks you if you want to run auto gift box or auto present box, I was finding myself having to switch back to ankulua frequently to switch which I wanted to run. 

If it works for you I should add a commit later, maybe add some settings so you can edit variables (like goldThreshhold and how many to check at a time) right in _auto_multi.

Inkaflare:
I gave it a try, it now works properly with a caveat; I was using gesture navigation on my Android over the classical 3 button layout at the bottom of the screen, and part of this setting is that to back out of fullscreen apps, you swipe upwards twice from the bottom edge of the screen (since the Home button is not there anymore at the right edge of the screen). When the script starts the scrolling motion, it seems to do so from that same very edge of the screen, so my phone recognizes it as the first swipe for minimizing the app instead and the scrolling isn't being performed.
When using classic three-button navigation, the script works fine and selects the correct embers as specified.
A simple fix for that was to edit the coordinates of where the swipe to scroll starts to a slightly higher position, replacing these lines
`		{ action = "touchDown", target = Location(700, 700) },
		{ action = "touchMove", target = Location(700, 175) },`
with these lines
`		{ action = "touchDown", target = Location(700, 650) },
		{ action = "touchMove", target = Location(700, 125) },`
Now it works for gesture system navigation as well.
Thanks a lot for the script!

reconman:
@Thelanator Thanks for your script, I worked on a version with a forgotten `Settings:set("MinSimilarity", 0.92)` so OCR didn't work.

From my side there are 2 improvements to make in your version of the script:
1. Change numberOCRNoFindException to numberOCR, otherwise it will treat failed OCR as the previous value and hide OCR problems.
2. Change the Locations of touchMove and touchUp at the bottom of the script to Location(700, 175). With those values, it will always scroll up exactly 4 entries instead of 3.9 entries.

potchy:
I can't seem to get the swipping in this script to be consistent enough. It seems that the game always scrolls up a bit further upon touch release. I've tried to tweak the parameters in many different ways, but it always gets offsync after a few swipes. jesus

@Inkflare makes a good point though

TheIanator:
Most of the for me it's able to scroll the whole list without becoming misaligned. It does get misaligned occasionally though, which is why I wrote it to search for check boxes and then calculate off of those instead of just checking where the text should be, worst case scenario it fails to check every fourth item, but when you hit receive selected everything gets realigned to try again. It's not perfect but as is has definitely saved me many hours.

Also it seems the values that got me exactly four scrolls almost every time weren't working for reconman, I wonder if mine works differently since my ankulua is using touch by accessibility instead of by daemon?

potchy:
The values that worked for me, with navigation gestured enabled, are:

```lua
local touchActions = {
	{ action = "touchDown", target = Location(700, 680) },
	{ action = "touchMove", target = Location(700, 130) },
	{ action =      "wait", target = 0.4 },
	{ action =   "touchUp", target = Location(700, 130) },
	{ action =      "wait", target = 0.5 } -- leaving some room for animations to finish
}
```

Can you test it to see if it breaks on your phone? I'd like to merge the script with these values, as it avoids the problem @Inkaflare mentioned.
I'm using root mode, which I think has the same speed as daemon.
I can't record a video right now, but at least for me accessibility mode has always been slow as hell. It takes like ~5 to complete a full swipe, while with daemon/root it takes ~1 second at most.

Inkaflare:
@potchy Tried on my phone; scrolling works without triggering the fullscreen exit gesture, but it scrolls a little over 4 items, resulting in the whole thing becoming misaligned after 4-5 scrolling motions and the script skipping every 4th item until the accumulated misalignment has skipped over one item entirely after a few more scrolling motions.

I wasn't aware that scrolling with daemon was that much faster btw. I thought that was an inherent limitation of Ankulua, lol. Sadly I have no other option than accessiblity as my phone is on Android 9 and the bootloader is locked.

TheIanator:
I had similar results, takes about two seconds to scroll with accessibility, scrolls a little too short and becomes misaligned around 10 scrolls or so.

Since JP works while USB debugging is enabled I did a little testing with both daemon and accessibility. I ran each a few times, but probably not enough to be truly scientific.

Daemon, New values, scrolls in 1 second, scrolls short, misaligned in 7 scrolls
Daemon, My values, scrolls in 1, scrolls short, misaligns in 3 or 4
Accessibility, New values, scrolls in 3 seconds, scrolls short, misaligns in 10
Accessibility, My values, scrolls in 3, scrolled through entire box (20~ scrolls) doesn't work flawlessly every time, tested 5 times, became misaligned twice, seemingly at random.

potchy:
135 worked great for me as well. Here's the demo:
https://youtu.be/ra98BT-QBx8

reconman:
I tested for 50 minutes and adjusting the Locations by suggested values and nothing really worked. Even when I subtract 50 from all Y values, it still scrolls too little. Make it scroll 5 pixels farther and it scrolls too far.

I'm using root clicking and my resolution is 1080p. Maybe that's a an error source? Maybe pixel positions are rounded up or down?

So far I've achieved the best results with the values from my previous comment.

reconman:
I forgot to mention that on my phone the OCR often detects cut off 3s at the top of the screen as < 3.

So that's why having accurate scrolling is important here.

TheIanator:
The region that is checked for checkmarks could be made smaller, so that the region checked for ocr is never partially offscreen. 

I think root/dameon is the more accurate method, and probably the one most people use, so probably have those set as the default and include a comment in the file and the readme that values may need to be tweaked on an individual basis for optimal performance.

Wild idea I had, but could a script be made that tests a lot of different scroll lengths and examines screenshots to generate optimal values for each users specific setup? Even if it's possible it's entirely possible that it's more trouble than it's worth.

potchy:
I downloaded your latest zip and updated your branch.
I guess we all think that this is good enough, so I'll be merging it as is.

Issue #327, opened 2020-01-18
Auto gift Exchange


FrayXQ:
is the auto gift exchange script supposed to reset boxes ?

TheRealPlagiariser:
Yes it will reset the box #329

reconman:
By adding some FinishedLotteryBoxRegion:highlight() calls I was able to find out that the coordinates of that Region is off. I was able to correct it by changing `local FinishedLotteryBoxRegion = Region(520, 840, 70, 100)` to `local FinishedLotteryBoxRegion = Region(575, 860, 70, 100)`

Should work with any region.

FrayXQ:
thanks, fixed it

reconman:
@FrayXQ Keep this issue open until they fix it.

reconman:
Duplicate of #303

PR #330, opened 2020-01-19
Fix region of finished lottery

reconman:
On the US server, the region for the greyed out "1" was way off. I don't know if this fix also works for non-US or other resolutions.

#327


potchy:
thanks

PR #335, opened 2020-02-11
Making some adjustments

TryBane:
Alright, this should be the definitive working version of what I had made before, with Preload NP. The name can always change, I still don't like it.

Starting off, this is only confirmed working in both JP and EN servers, in all aspects I'm about to list. To get it working in the other regions, I'll need screenshots of the servers in question in these screens:

Support screen
Withdraw screen (on failed mission where you choose to withdraw, spend Saint Quartz [Stones], or use Command Seals
Connection Failed screen

With these 3 screenshots from TW and CN servers, I can make it function 100%.

So to list the new features.

Firstly, Preload NP is working and is better than before! Before it would spring up toast messages about being unable to read cards, but now that is no longer an issue! The cards are all being read and cached in a list locally within card.lua, which is then reset to a clear list after battle.lua finished performBattle. I have adjusted a bit in card.lua to make this work, but essentially there are 2 new functions, readCommandCards and resetCommandCards. readCommandCards stores the array locally, and which can be called by battle.lua. Thinking back, I believe this is possible to do without adding this function, but I didn't want to damage anything that wasn't broken, so this option was the safer choice. Then with resetCommandCards, the list of command cards stored is reset to a blank list, and then "cardsClickedSoFar" is reset to 0. cardsClickedSoFar is a new variable added for the PreloadNP feature, which keeps track of how many cards have been clicked. In the clickCommandCards function, this value is checked to ensure any cards that are already clicked are not clicked again.

There is a small added portion within clickCommandCards that I have that is the beginning of my attempts to get chaining to work. This has no influence on how things work now, so it can be ignored. I made sure the file would work regardless of updated config files. These changes can be found in the battle.lua in the init function, where it checks if "chains" is nil (which would be in the config file) and sets chaining to false if it is the case. Thus, nothing changes in regards to the function of the script.

A change to Support, I changed the scrolling method from the manualTouch option to swipe. This adjustment was made in order to increase the speed of scrolling. The swipe can be adjusted as needed for those that wish to, but I made it small enough that slower devices should still be able to use it without issues. This can be adjusted as we get feedback. The location of the change to occur from feedback is line 168 and 169 of support.lua. 169 is the wait time to wait for scrolling to finish, and 168 is how fast it scrolls. Smaller is faster for line 168. If people find the scrolling to skip servants, increase this number. If people find that it starts scrolling again before it finishes the movements, increase the wait time on 169.

The new features Withdraw and Retry are working on JP and EN servers! I acquired the necessary screenshots in order to crop the buttons needed for these features to work. The Retry feature works on both, and Withdraw has at least been tested on EN servers (haven't tried it on JP. Please give some feedback on whether it works or not!) To test it, or acquire screenshots, simple enter one of the Fuyuki quests (I did the last one for 7 Stamina/AP) with lvl 1 servants. A level 40 Mashu will be able to clear the quest, so if you don't want to waste the Stamina/AP, choose a lower level support. Fail the quest, and take a screenshot where it says "Withdraw" as an option.

One thing to note, and I took this into consideration when writing this feature in, based on what is available, the "Withdraw" button will move. There can be from 1 to 3 buttons available. If you don't have 3 Command Seals, or no Saint Quartz (Stones) then it'll be decreased by 1 or 2 if both are unavailable. To compensate for this, I adjusted the method of choosing the click location.

If you take a look at the regular.lua file, you'll see on line 26 a new field: MatchClick. This is to utilize an optional clicking method that we haven't been using in the script up to this point. It is used in both Retry function and Withdraw function located on lines 72 and 83 respectively. MatchClick, in their Validator functions, gets set to the returned value of the region:exists function in each of the Validator functions. This is then used as the click location in the respective Actor functions. Afterwards, this field is set to nil to prevent false positive clicking. If MatchClick is set to nil, and the function attempts to click the nil value, the script will crash. This shouldn't happen, as I should've been successful in keeping such a thing from occurring, but in the case that it does, we'll need to double check the functions associated with MatchClick (only Withdraw and Retry for now).

Now, Retry is a feature many have been asking for. It reads the menu in which states that connection couldn't be established or was lost. And then proceeds to click the button to retry to connect. I have done what I could to account for all instances where this occurs and my best option was to list it with the Validators and Actors loop. In this way, it'll be checked on every iteration of the loop and should account for most cases where it occurs. Note that this has to be the FIRST Validator/Actor option, otherwise it'll read the other menus first and fail to recognize it has lost connection (this gave me a huge headache to figure out by the way).

Now there was at least one instance where this didn't account for connection problems to occur, which was updated friends list on support selection. Because the support selection is a closed loop, once it enters it, it won't leave until a support is selected. Thus, it'll never re-enter the Validator loop and never see that Retrying was needed. This wasn't necessarily a problem though, as in the process of pressing the update button, it would end up clicking the retry button anyways. However, this could still cause problems in certain instances, thus I came up with a solution. Said solution was a loop within the selectPreferred loop located on line 152 of support.lua. This loop will check if Retry is needed and then click Retry when it sees it. Then loop again and again as necessary (some connection issues can span for minutes if you are in a bad area). At least on my own device, the wait time located on line 75 of regular.lua was more than plenty between Retry attempts.

If you guys have any feedback and suggestions, feel free. I'll implement any changes that are needed as soon as I have time. Alternatively, feel free to make any changes yourself if you think it is simple enough.


TryBane:
Also! Forgot to mention this, but I also resolved the issue in regards to the "support_region_tool" not working as initially intended! I just remembered this when clearing up the conflicts listed for the merge. Essentially changed the tool from the quick card to the "confirm support setup" button. This required a small adjustment to the offset from 156 to 114 due to the location of the button. This also resulted in a changed to the Region used for the region anchor. Because of this change, there should be no conflicts resulting in having Command Codes applied to support servants!

FGO_EN_REGULAR.lua
```diff
@@ -34,6 +36,7 @@ Support_FriendsOnly = 0
 Support_FriendNames = ""
 Support_PreferredServants = "waver4.png, waver3.png, waver2.png, waver1.png"
 Support_PreferredCEs = "*chaldea_lunchtime.png"
+Support_FriendNames = ""
```
potchy: this is duplicated

modules/battle.lua
```diff
@@ -82,16 +100,22 @@ performBattle = function()
...
 	if UnstableFastSkipDeadAnimation == 1 then
 		skipDeathAnimation()
 	end
-
-	wait(2)
+	
+	_card.resetCommandCards()
+	
+	if NPsClicked then
+		wait(25)
```
potchy: why this hella long delay?

modules/support.lua
```diff
@@ -150,6 +159,9 @@ selectPreferred = function(searchMethod)
 			click(game.SUPPORT_UPDATE_CLICK)
 			wait(1)
 			click(game.SUPPORT_UPDATE_YES_CLICK)
+			while NeedsToRetry() do
```
potchy: I think the code for retrying should be moved to a different module, unless you remove `local` from the function declaration, but that would look dirty.

image

modules/support.lua
```diff
@@ -163,17 +175,8 @@ selectPreferred = function(searchMethod)
...
-		{ action = "touchMove", target = game.SUPPORT_SWIPE_END_CLICK },
-		{ action =      "wait", target = 0.4 },
-		{ action =   "touchUp", target = game.SUPPORT_SWIPE_END_CLICK },
-		{ action =      "wait", target = 0.5 } -- leaving some room for animations to finish
-	}
-
-	-- the movement has to be as accurate as possible
-	setManualTouchParameter(5, 5)
-	manualTouch(touchActions)
+	swipe(game.SUPPORT_SWIPE_START_CLICK,game.SUPPORT_SWIPE_END_CLICK,.5)
```
potchy: I also tried using `swipe` when I originally wrote this function, but unfortunately it doesn't work for all devices, so I used `manualTouch` for better compatibility. Here's what happens on my phone: https://www.youtube.com/watch?v=X4eKXNaOyjk

TryBane:
Hmmm, okay. So I'll go in order here.

My bad on the duplicate. I didn't realize that that was there XD

The delay is for NP Animation. 25 seconds is about average NP duration length. I specifically made it so that it wouldn't hit this length if an NP wasn't touched. It also only accounts for 1 NP animation. But basically this delay allows the pone to not be constantly searching for images, which can save on battery life and causes less lag. I can't 100% confirm it either, but I also think it lowers the rate which Accessibility issues arise too.

I tested the retry function in Support menu and it worked fine for me though.... Not sure what the problem is on that? I might have to take another look. Otherwise, just making it a global as you mentioned, or writing a duplicate in the module itself. Both should work.

I really wish Swipe worked more globally.... I've seen people have issues with Manual Touch though too so I'm not sure what to do about this. Swipe is definitely faster between the two though.

modules/support.lua
TryBane: Alright, code has been reverted back to the Manual Touch function.

modules/support.lua
TryBane: Not sure what the issue was to be honest. It worked fine for me when I tested it myself. But regardless, I made some changes so that it is in now in the Game module as a global function. Check it out and see if it works now.

modules/battle.lua
TryBane: NP animations can be really long, but this isn't the longest delay that could be used for them. I made it slightly shorter than the average I believe.

TryBane:
Oh shit, thanks for catching that! I didn't test it after making the change. Sorry!

reconman:
Next time I would recommend squashing the commits into 1 so master isn't flooded with commits like "Reverting X", "Fixing X", "Restore accidentally deleted..." like it's the case here.

There should be a squash button on every open Pull Request, here's some info https://github.blog/2016-04-01-squash-your-commits/

TryBane:
Thanks for this. I've only worked on a few projects that revolve around Git so I never had the chance to learn about this. Glad I know it now, it'll be nice for future pull requests!

reconman:
It was more directed towards @potchy in this case, the one merging the Pull Request has to use that button.

TryBane:
Yeah, I didn't realize that until after I read the link you posted (commented before checking it). But still appreciate it myself as well. It'll be useful for future projects.

reconman:
There's also an option to only allow Squash Merges in the repository settings, only @29988122 can do that. The option can be found in https://github.com/reconman/Fate-Grand-Order_Lua/settings under "Merge Button".

@29988122 Could you deactivate "Allow rebase merging" and "Allow merge commits"? Then every Pull Request only produces a single commit where the merger has to choose the combined commit message. That's how we do it in the company project and it works really well.

potchy:
I appreciate your concern, but it was a conscious choice. I usually rebase to maintain a linear history, but since there were many commits this time around, I thought it was a good idea to group everything together. It's possible to compare merge commits as a whole anyway. In retrospect, this PR's original messages had a lot of "fixed this", "fixed that", "oops", so I agree a squash would have been better. 🤔

Issue #101, opened 2018-11-22
Add auto support selection to autoskill lists.


5urmount:
Make it so that we can choose one list during the menu which contains both auto support and autoskill, so that I can run separate lists that farm for exp (searches for chaldea lunchtime) and qp (searches for mona lisa) in the same script.

TryBane:
I was literally about to add this suggestion. I'll take a look at the auto-skill code that uses a list and see about implementing something similar for the support select. Shouldn't be too complicated.

29988122:
@TryBane 
If you're going to touch that, feel free to modify the behavior between BOTH autoskill AND autoskill_list.

I was writing for myself so the implementation was rather quick and dirty.

TryBane:
I have a version of the enhancement ready. Just need to figure out how to submit it as a pull request?

First time submitting a pull request to a git that wasn't my own.

TryBane:
I can always adjust it a different way too. Here I linked the support and CE selection to the Auto Skill selection used. But I can also allow you to list Support and CE selections that you can choose independently from one another.

For instance, say you have Waver in your support selection. And right now you're farming for the Cheese Cakes for the NA event currently ongoing. I could make it so that instead of making a completely new support selection, you will have another dialog box open asking you to select support, then CE. It would be more modular, expanding scalability of this script.

5urmount:
One thing I think you should consider is an option to run indefinitely, so once you pick a list, it doesn't keep asking you every instance refresh.

TryBane:
Hmmm, it shouldn't be asking every instance refresh. Only when you first
initiate it.

5urmount:
Weird, might have ran the wrong script. Will report back when enough AP has regenerated.

TryBane:
Yeah no problem. If there is an issue, then give me the details of what is
happening. I'm running it right now, specifically running skill set 7.

Give any details such as how many swipes you set it to, what type of
support (first, preferred, etc) and when it actually happens like every
run, whenever you run out of AP, etc.

On Nov 29, 2018 23:06, "5urmount" <notifications@github.com> wrote:

Weird, might have ran the wrong script. Will report back when enough AP has
regenerated.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/29988122/Fate-Grand-Order_Lua/issues/101#issuecomment-443084198>,
or mute the thread

MathewSachin:
[...]

PR #349, opened 2020-03-21
Common startup file for all game server regions

MathewSachin:
Currently, different startup files are used for different regions as `FGO_XX_REGULAR.lua` where XX can be CN, JP, TW or EN.
I found that pretty much all the code in these files is the same except 3 variables, namely `GameServer`, `SupportSwipeEndClick` and `StageCountRegion`.

So, I've created a single file where you just need to set `GameServer` variable.
`SupportSwipeEndClick` and `StageCountRegion` are set automatically in the `game.lua` module based on the `GameServer` value.
I'm not sure what to name the file, so I put it as `FGO_REGULAR.lua`.
I've updated `README.md` and comments in code where the file name was referred.
The wiki would also need to be updated if this is to be accepted.

This would make it easier to add new options or change existing ones without modifying multiple files.
Let me know what you all think about this.

Based on the discussion in #101, this would help if my suggestion of mixing AutoSkill with Support selection is accepted.


FGO_REGULAR.lua
```diff
@@ -1,105 +1,103 @@
---Internal settings - do not modify.
```
reconman: Please convert all Linux style endings (LF) to Windows style (CR LF)

MathewSachin: I double-checked and it seems that it is already in CRLF.
If you're wondering why git diff isn't showing properly, it should be because the file name was changed.

reconman: Actually, it is LF on your repository. I downloaded the file from your repository and it didn't have CRs.

I think you have `core.autocrlf` activated for your repository, which converts all CRLFs into LFs when you commit the files and converts LFs into CRLFs when you load files from the repository.

Disable this by executing `git config core.autocrlf false` in your repository folder.

MathewSachin: Yup! it's set to true, so that's the problem. I'm sorry.
I'll commit changes in a few minutes

z0rc: @reconman asking every contributor to change default git settings is meh. Instead you can set up expected line ending in `.gitattributes` here in repo. See https://help.github.com/en/github/using-git/configuring-git-to-handle-line-endings#per-repository-settings

MathewSachin: I've changed line endings to CR LF and rebased into a single commit.

@z0rc That would be pretty nice.

reconman: @z0rc You're right, I will set it up. I actually read that article shortly before you linked it.

@MathewSachin Sorry, I messed up in my own PR, where autocrlf was also enabled. Since there's no option to convert all line endings to CRLF on committing, I will instead force everyone to commit LFs.

MathewSachin: No problems, it happens sometimes, fine then I'll merge your changes from master to here and make sure to change back to LF

reconman: @MathewSachin All files in the repository now use LF and are converted to CRLF if you're on Windows.

Now your branch has a merge conflict with master, because you want to commit CRLF endings. I'm very sorry for wasting your time.

reconman: @MathewSachin Btw, you can convert all line endings automatically by running `git add --renormalize .`

MathewSachin: Well, I already did that manually 😆 
Now, there are many commits and I'm unable to rebase correctly.
If anyone merges this PR, it would be best to use the Rebase/Squash and merge option.

reconman:
I merged your changes into master.

MathewSachin:
Cool!
Are there any updates needed in the wiki? (I'm only familiar with English)

potchy:
Good idea

PR #351, opened 2020-03-22
Add AutoSupportSelection to AutoSkill list

MathewSachin:
Requested in #101

I've added a `middleware.lua` file and moved the `PSADialogue` function into it.

The flow is like: `FGO_REGULAR.lua -> middleware.lua -> regular.lua`

`Autoskill_List` array is now defined by the user instead of being restricted to 10 items.
Each entry should have `Name` and `Skill_Command` members.
Optionally, `AutoSupportSelection` options like `Support_SelectionMode`, `Support_SwipesPerUpdate`, `Support_MaxUpdates`, etc can be used to override the corresponding options defined outside the AutoSkill list.

I removed the `Enable_Autoskill_List` variable. I think `Enable_Autoskill` should be sufficient alone.

### DIFF
--- a/FGO_REGULAR.lua
+++ b/FGO_REGULAR.lua
@@ -1,19 +1,4 @@
---Internal settings - do not modify.
---***************************************************************************
-dir = scriptPath()
-setImagePath(dir)
-
 Debug_Mode = false -- set to 'true' if needing to debug
-
---Initalize for user input listnames
-Autoskill_List = {}
-for i = 1, 10 do
-	Autoskill_List[i] = {}
-	for j = 1, 2 do
-		Autoskill_List[i][j] = 0
-	end
-end
---***************************************************************************
 
 -- Can be EN, JP, CN or TW
 GameRegion = "EN"
@@ -25,7 +10,7 @@ Refill_Enabled = 0
 Refill_Resource = "All Apples"
 Refill_Repetitions = 0
 
---AutoSupportSelection
+--AutoSupportSelection Defaults
 Support_SelectionMode = "first"
 Support_SwipesPerUpdate = 10
 Support_MaxUpdates = 3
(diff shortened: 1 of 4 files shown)


MathewSachin:
@reconman @potchy Should we include `Battle_CardPriority` in autoskill list too?

potchy:
Yes, I think you should include all options, actually!

I have an idea to solve that bunch of ifs in middleware.lua.
It isn't pretty and I doubt it's recommended by any serious Lua programmer, but it works.
```lua
-- PSADialogue()
selected_autoskill = {
	Support_SelectionMode = "preferred"
}

-- SetSupportOptions()
Support_SelectionMode = "manual" -- this is a default value
Support_MaxUpdates = 10

for key, defaultValue in pairs(_G) do
	userValue = selected_autoskill[key]
	
	if userValue ~= nil then
		_G[key] = userValue
	end
end

-- this is here just for the sake of a demo
for key, currentValue in pairs(_G) do
	print(key, ": ", currentValue)
end
```

Result:
```
getmetatable	: 	function: 0x41c0b0
tostring	: 	function: 0x41b480
print	: 	function: 0x41b8e0
utf8	: 	table: 0x12909f0
rawset	: 	function: 0x41b7b0
table	: 	table: 0x128ddc0
assert	: 	function: 0x41c180
selected_autoskill	: 	table: 0x128f1a0
error	: 	function: 0x41bba0
os	: 	table: 0x128dd30
load	: 	function: 0x41bf20
io	: 	table: 0x1292710
select	: 	function: 0x41b4b0
pcall	: 	function: 0x41b420
setmetatable	: 	function: 0x41bd00
xpcall	: 	function: 0x41b320
string	: 	table: 0x128f350
coroutine	: 	table: 0x128da60
_VERSION	: 	Lua 5.3
next	: 	function: 0x41ba10
rawequal	: 	function: 0x41b8a0
------------> Support_MaxUpdates	: 	10 <------------
rawget	: 	function: 0x41b800
------------> Support_SelectionMode	: 	preferred <------------
bit32	: 	table: 0x128b9a0
debug	: 	table: 0x128d640
math	: 	table: 0x128fb10
pairs	: 	function: 0x41be70
ipairs	: 	function: 0x41be90
rawlen	: 	function: 0x41b850
_G	: 	table: 0x128b930
tonumber	: 	function: 0x41b570
type	: 	function: 0x41b3b0
collectgarbage	: 	function: 0x41bc30
```

Currently I don't have to test it myself in the script, but I hope it was of some help.

MathewSachin:
@potchy This method is super good. I'll give it a try.
Honestly, I was tired of seeing those many `if`s too. This really helps.

MathewSachin:
`Name` is the only variable we don't want to override. So, we can write it this way without checking the global value:

```lua
for key, value in pairs(selected_autoskill) do
	if key ~= "Name" then
		_G[key] = value
	end
end
```

This also works for cases where the user didn't define the variable globally.
Options like `BoostItem_SelectionMode` are not required to be defined, the code checks whether they are defined before accessing them.

https://github.com/29988122/Fate-Grand-Order_Lua/blob/c716d897945932e6eb8e326775ac3db65ce31c33/regular.lua#L89-L92

MathewSachin:
I am thinking of one more idea from this method.
Instead of checking if the user has defined the variable or not like it is done for `BoostItem_SelectionMode`. We could keep a `Defaults` table and fill missing values from it.
So, the other modules can access those variables without worrying about them being `nil`.

```lua
local Defaults = {
    BoostItem_SelectionMode = "disabled",
    Story_Skip = 0,
    Debug_Mode = false,
    StopAfterBond10 = 0,
    UnstableFastSkipDeadAnimation = 0
    -- and more
}

for key, defaultValue in pairs(Defaults) do
	globalValue = _G[key]
	
	if globalValue == nil then
		_G[key] = defaultValue
	end
end
```

We could even remove these options from the starter script and put them in documentation to reduce clutter.
Yeah, but, this should ideally be done in a different pull-request.

MathewSachin:
@potchy @reconman 
Do you want me to make any other changes?
I've tested the script and it's working fine for me.

potchy:
Thanks!
I tested it for a bit and it seems to be working great. 😄 
Please help other users if they encounter problems after this change.

MathewSachin:
On this wiki page: https://github.com/29988122/Fate-Grand-Order_Lua/wiki/Script-configuration-English, the old configuration is still mentioned. It can be redirected to the README, like what is done for AutoSupportSelection.



> I am thinking of one more idea from this method.
> Instead of checking if the user has defined the variable or not like it is done for `BoostItem_SelectionMode`. We could keep a `Defaults` table and fill missing values from it.
> So, the other modules can access those variables without worrying about them being `nil`.
> 
> ```lua
> local Defaults = {
>     BoostItem_SelectionMode = "disabled",
>     Story_Skip = 0,
>     Debug_Mode = false,
>     StopAfterBond10 = 0,
>     UnstableFastSkipDeadAnimation = 0
>     -- and more
> }
> 
> for key, defaultValue in pairs(Defaults) do
> 	globalValue = _G[key]
> 	
> 	if globalValue == nil then
> 		_G[key] = defaultValue
> 	end
> end
> ```
> 
> We could even remove these options from the starter script and put them in documentation to reduce clutter.
> Yeah, but, this should ideally be done in a different pull-request.

What do you think about this?

Issue #350, opened 2020-03-22
Withdrawing is not configurable


Rajvir8:
For the first time I noticed my auto withdraw after losing a battle rather than getting stuck on the screen, I just thought I'd ask if that was a new feature that was added.  If so is there a way to easily turn it back off?  Generally speaking it's best to be able to just use a CS if I've lost as I rarely need them.

PR #352, opened 2020-03-22
Made withdrawing configurable

reconman:


reconman:
Fixes #350

potchy:
thx!

reconman:
I just noticed that 0 and 1 are not true and false, so my if condition doesn't work. Why are so many _Enabled variables set to 0 and 1?

potchy:
the code base already followed this pattern when I joined
I don't mind it personally

PR #355, opened 2020-03-24
Fix supportBound region, correct Merlin images

reconman:
I replaced the MLB image with one I took myself and now support selection is much more reliable. Can anyone check if it also works for other people?


reconman:
Doesn't really look perfect, it still fails plenty of times.

potchy:
At the time I developed this feature there wasn't transparency support.
Maybe it would work better if you used [mask()](https://sikulix-2014.readthedocs.io/en/latest/pattern.html#Pattern.mask) in combination with the whole MLB star in a .png with transparent background.

reconman:
In most cases the script can't detect Merlin, where I've noticed that all the Merlin images don't have MLB icons on them. merlin1.png and merlin23.png include areas that are not visible if Merlin is max ascended.

I'll update the Merlin images soon and then check again if the old MLB image causes any problems.

reconman:
So, the true perpetrator was the region where the support module looked for the servant and CE images.

The Y value was too low, so part of the CE was always cut off.

I checked the position of supportBound with highlight(). It definitely works in NA and should work in JP. (only 1 pixel difference in the size of support_region_tool.png)

reconman:
@potchy I'm finished with my PR unless you can find a PNG of the MLB star lying around somewhere. I was not able to find any image dump or unencrypted APK on the internet and extracting the full MLB star from a screenshot is hard.

Please use squash and merge if you dedice to merge it, because I reverted the new MLB image.

potchy:
merlin23.png isn't working for me (Nox 1280x720)
It's looking very disproportional compared to the previous one, to be fair. merlin1.png too.
You sure you took them from screenshots in 16:9 aspect ratio?

image

reconman:
I'll check tomorrow, this time I'll use Photoshop on my PC.

potchy:
Can you let me take a look at the original screenshot you took?

reconman:
https://dl3.pushbulletusercontent.com/LXNTFYokFRWqpxkDMcXs5CC56PfPFnzF/Screenshot_20200329-143413.jpg
https://dl3.pushbulletusercontent.com/cZ4TaX7BRAJIrG3hSre8EW2lu9nrNTKf/Screenshot_20200329-143339.jpg

I downscaled the screenshots to 720p, then extracted the images.

PR #356, opened 2020-03-29
Default values for options

MathewSachin:
This PR adds code for automatically setting defaults, if the user did not specify some options in their script. This would allow to reduce clutter in the script, the user would only need to specify the options they want.
I've removed a few options from the `FGO_REGULAR.lua` file and wrote about them in `README.md`.

Instead of checking if the user has defined the variable or not like it is done for `BoostItem_SelectionMode`. We could keep a `Defaults` table and fill missing values from it.
So, the other modules can access those variables without worrying about them being `nil`.

```lua
local Defaults = {
    BoostItem_SelectionMode = "disabled",
    Story_Skip = 0,
    Debug_Mode = false,
    StopAfterBond10 = 0,
    UnstableFastSkipDeadAnimation = 0
    -- and more
}

for key, defaultValue in pairs(Defaults) do
	globalValue = _G[key]
	
	if globalValue == nil then
		_G[key] = defaultValue
	end
end
```

### DIFF
--- a/FGO_REGULAR.lua
+++ b/FGO_REGULAR.lua
@@ -51,6 +51,18 @@ Battle_AutoChooseTarget = 1
 --NoblePhantasm Casting
 Battle_NoblePhantasm = "disabled"
 
+-- set to 'true' if needing to debug
+Debug_Mode = false
+-- stop the script after retreiving a Bond 10 CE card
+StopAfterBond10 = 0
+-- Boost item, possible values: disabled, 1, 2 or 3. Make sure "Confirm Use of Boost Item" is off
+BoostItem_SelectionMode = "disabled"
+StorySkip = 0
+--Auto Withdrawing
+Withdraw_Enabled = false
+--FastSkipDeadAnimation
+UnstableFastSkipDeadAnimation = 0
+
 -- Do not modify below this line
 dir = scriptPath()
 dofile(dir .. "middleware.lua")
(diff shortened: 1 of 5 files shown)


potchy:
I like the defaults table thing, but I don't think it's a good idea to remove those options from the configuration file altogether. I'm afraid it would make them too obscure or something. @reconman what about you?

reconman:
I have a similar opinion as potchy regarding removing stuff from the regular.lua file. Users don't like to read instructions, they just look at the file.

If we remove almost everything from the regular file, people will be discouraged by the giant README and will never discover certain features or will create lots of feature requests for things we already have.

What we could do to reduce the size of the script is to remove most comments and document everything well in the Github project instead. That way people will notice the variables, wonder what they do, and then consult the documentation.

MathewSachin:
I definitely agree with you too.
I'd add the variables back to the script like @reconman suggested, without the comments.
So, people who are curious about the feature can consult the README.

Or, should we add documentation outside the README, like more markdown files within the repo itself and point to them from README. This can prevent the README from becoming too big.
GitHub wiki is great and all but outside contributors can't send pull-requests there.
And, it can't be updated side by side with the code.

MathewSachin:
I've put those variables together at the bottom of the script with single-line comments briefly describing their use.

potchy:
@reconman feel free to merge stuff if I'm not around

reconman:
@potchy Haven't looked at the PR in detail yet, that's why I didn't merge it

PR #361, opened 2020-04-12
Add Support Image Maker script

MathewSachin:
I've added a script that can make images that can be used with Auto Support Selection.
The generated images are black and white, but that is not a problem because AnkuLua matches images in black and white.

I've added the scaling script into it so I think it should also work fine on phones with a notch too.

To run the script, select it in AnkuLua, go to support selection screen and press Play.
The script ends almost immediately.
For images to be saved, the Servant+corresponding CE should be completely on screen.
Only 2 Servant+CE can be completely on screen at a time.
The images are stored in the same directory as the script and named as: `ce_0.png, ce_1.png, servant_0.png, servant_1.png`.

Some generated images:

### DIFF
--- a/_support_img_maker.lua
+++ b/_support_img_maker.lua
@@ -34,7 +34,7 @@ for _, testRegion in ipairs(regionArray) do
   supportBound:setY(testRegion:getY() - 70 + 68 * 2)
 
   if ankuluaUtils.DoesRegionContain(screenBounds, supportBound) then
-    local servantBound = Region(supportBound:getX(), supportBound:getY(), supportBound:getW(), 88)
+    local servantBound = Region(supportBound:getX(), supportBound:getY(), 250, 88)
     servantBound:save('servant_' .. i .. '.png')
 
     local ceBound = Region(supportBound:getX(), supportBound:getY() + 160, supportBound:getW(), 60)
--- a/_support_img_maker.lua
+++ b/_support_img_maker.lua
@@ -51,5 +51,5 @@ end
 if i == 0 then
   scriptExit("No support images were found on the current screen")
 else
-  scriptExit(i .. " servant and ce images stored into '" .. folder .. "' folder")
+  scriptExit(i .. " Servant and CE images have been stored in the  folder '" .. folder .. "'")
 end
--- a/_support_img_maker.lua
+++ b/_support_img_maker.lua
@@ -1,0 +1,45 @@
+dir = scriptPath()
+package.path = package.path .. ";" .. dir .. 'modules/?.lua'
+setImagePath(dir)
+
+local ankuluaUtils = require("ankulua-utils")
+local scaling = require("scaling")
+
+GameRegion = "EN"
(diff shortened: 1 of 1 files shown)


potchy:
Amazing. :o
My only nitpick is that I think the snapshots should be saved to a different folder.

MathewSachin:
Yeah, I was thinking of that too but couldn't settle on a location. Open for suggestions.

reconman:
Since you would normally execute this script a few times in a row to collect different CE images, I would suggest creating a folder per script run with a folder name containing the current date and time.

I don't know if Ankulua supports creating folders, though.

My idea is to use something like this:
```
--gets current date and time in the format 20200412-221630
local datetime = os.date('%Y%m%d-%H%M%S')
local folder = 'support_' .. datetime
region:save(folder .. '/servant_' .. i .. '.png')
```

If Ankulua can't handle creating folders through save, you could also try `os.execute('mkdir ' .. folder)`

reconman:
Also we could make the servant images less wide so the MLB star is automatically cut off. So instead of `supportBound:getW()` use `125`

MathewSachin:
@reconman I've made the updates you suggested.
Do let me know if there is any other change to make.

MathewSachin:
The wiki page: [How to take a screenshot and resize properly](https://github.com/29988122/Fate-Grand-Order_Lua/wiki/How-to-take-a-screenshot-and-resize-properly), has to be updated with info about how to use this script.

_support_img_maker.lua
```diff
@@ -0,0 +1,55 @@
...
+
+    local ceBound = Region(supportBound:getX(), supportBound:getY() + 160, supportBound:getW(), 60)
+    ceBound:save(folder .. 'ce_' .. i .. '.png')
+
+    i = i + 1
+  end
+end
+
+if i == 0 then
+  scriptExit("No images generated")
```
reconman, line 52:
```suggestion
  scriptExit("No support images were found on the current screen")
```

_support_img_maker.lua
```diff
@@ -0,0 +1,55 @@
...
+else
+  scriptExit(i .. " servant and ce images stored into '" .. folder .. "' folder")
```
reconman, line 54:
```suggestion
  scriptExit(i .. " Servant and CE images have been stored in the  folder '" .. folder .. "'")
```

reconman:
@MathewSachin @potchy Regarding the wiki changes, I'm currently creating a better documentation using Github pages + [docsify](https://docsify.js.org/). I'll create a Pull Request for that once I'm done.

Then everyone can suggest changes to the documentation.

reconman:
@MathewSachin Do you mind adding documentation to https://29988122.github.io/Fate-Grand-Order_Lua/#/support-image-maker (support-image-maker.md) and also linking to it from the Extra Scripts page?

Right now the page is referenced by features.md at the "Follow this guide" part.

MathewSachin:
Sure!

PR #363, opened 2020-04-14
Document Support Image Maker script

MathewSachin:
Hope I did it right. Let me know if there are any changes to make.

@reconman

### DIFF
--- a/docs/extra-scripts.md
+++ b/docs/extra-scripts.md
@@ -36,3 +36,9 @@ The configuration parameters are:
 | Max_Click_Count | The maximum number of presents to collect.<br><br>If your inventory space is very limited, you should choose a low number.                                                                             |
 
 You will probably have to optimize the Y positions in the `touchActions` part of the script yourself, since scrolling behaves differently dependening on your device and Ankulua settings.
+
+## `_support_img_maker.lua`
+
+You can use this script to make your own images to be used with `Auto Support Selection`.
+
+See the guide [here](support-image-maker.md).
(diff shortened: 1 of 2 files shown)


reconman:
@MathewSachin I've made some small corrections, thank you very much for your work so far.

MathewSachin:
Well, I'd like to thank you all for your work on this script too.
Really eases the grind.

I also tried porting the whole script along with features of AnkuLua as an app too: https://github.com/MathewSachin/Fate-Grand-Automata. Didn't want to advertise it or anything. The app I made works fine for me but might not be good enough for general use. And it's a pain to debug.

Just wanted to tell you that Support Image Maker came from there.
Still, the credit is to you guys for your awesome work on this script.
I'd try to keep contributing whenever I have the opportunity to.

MathewSachin:
Also, let me know if there are any features I can try working on. I can test only on NA version but I want to help if I can.

reconman:
FGO automation without a paid subscription sounds nice, and I like C# much more than Lua, although I don't have much experience with C# and no experience with Android apps.

Too bad the Ankulua core scripts are closed source, so you have to guess how they've implemented new features.

I'll probably take a look at your app soon.

There's one thing I wanted to mention regarding this feature. I hoped it would work in the Friend List, but sadly it's off by a few pixels for both the Servant and CE images.

Since the Friend List screen probably looks different on JP than on NA, I gave up on correcting the script. (Let the user decide between Support Selection and Friend List)

MathewSachin:
We might be able to check which screen we are on, whether we are on support selection screen or on friend list screen and then use the corresponding region.
I'll give this a try later and let you know.

PR #364, opened 2020-04-15
Support Image Maker script can be used from Friend List

MathewSachin:
Making the `_support_img_maker.lua` script usable on Friend List.
Also, adjusted the height of the generated image for CEs to remove MLB star.

~~I'll mark as *Ready for Review* once I'm done testing.~~

### DIFF
--- a/_support_img_maker.lua
+++ b/_support_img_maker.lua
@@ -4,6 +4,7 @@ setImagePath(dir)
 
 local ankuluaUtils = require("ankulua-utils")
 local scaling = require("scaling")
+local game = require("game")
 
 GameRegion = "EN"
 GeneralImagePath = "image_" .. GameRegion .. "/"
@@ -17,6 +18,9 @@ scaling.ApplyAspectRatioFix(SCRIPT_WIDTH, SCRIPT_HEIGHT, IMAGE_WIDTH, IMAGE_HEIG
 
 -- Single screenshot is enough
 snapshot()
+
+--Checks if Support Selection menu is up
+local isSupportScreen =	game.SUPPORT_SCREEN_REGION:exists(Pattern(GeneralImagePath .. "support_screen.png"):similar(.85))
 
 local supportBound = Region(106,0,286,220)
 local regionAnchor = Pattern(GeneralImagePath .. "support_region_tool.png")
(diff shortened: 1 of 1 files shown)


MathewSachin:
I've tested it and now it works from both Support Selection as well as Friend List (at least on NA it does).

MathewSachin:
@reconman

29988122:
Guess this one's on me.

Will test and confirm the result here within 24~48hours.

29988122:
1. please do some exception handling when support_region_tool.png not found on screen.
(for example, wrong region setting, wrong screen..)

2. Remind player to switch region settings in the script, cause this one's under root folder;
or try to read the region variable inside FGO_REGULAR.lua;
or just pop a toast and tell them to RTFM(read the fucking manual).

Saw too many tickets with error java io exception. lol.

3. MLB star not included in the snapshot - I know it's intended. 

Otherwise fine, working very well.
Tested on emulator 1280*720 JP version. Too lazy to test other combinations. XD
Thank you *very much* for your contribution!

reconman:
@29988122 So what do you want exactly regarding 2. and 3.?

How should the reminder look like? Why do you want the MLB star in the screenshot?

MathewSachin:
@reconman Regarding 1, I think changing the message when no images are generated should be sufficient, right?

We can tell the user to `Make sure you're either on Support selection or Friend list screen. Also, see that the 'GameRegion' parameter in this script is set correctly.`
I think this can be worded better though.

reconman:
Okay, so I would change https://github.com/29988122/Fate-Grand-Order_Lua/pull/364/files#diff-70c53157a32dd4e6f12107acf0ef0e42L52 to `No support images were found. Make sure you're in the Support Selection or in the Friend List and that the correct GameRegion is set in the script.`

I wouldn't link the GameRegion setting to the other scripts since you can never know whether a user has renamed  their files.

29988122:
> 
> 
> @29988122 So what do you want exactly regarding 2. and 3.?
> 
> How should the reminder look like? Why do you want the MLB star in the screenshot?

My wording's bad. Sorry.
1 and 2 are probably the same issue. Due to the fact that I did not change the gameregion variable to JP, and IO error popped. 
We gotta find a way to solve this(via reminder or anything), since user probably won't figure that out, and they just give up or write issue ticket.

3 is just to make sure this behavior is intended, since the old behavior was to includ the MLB star for support_selection module to use. (probably deprecated behavior, that's why I asked)

reconman:
What kind of IO Error happened? Can you provide a screenshot?

Regarding the MLB star on the CE, it's better to store CE images without a star and use the existing isLimitBroken function in support.lua. I've successfully used the images produced by the script in combination with restricting it to MLB.

29988122:
圖片
The chinese reads: unable to open /sdcard......

After I changed region from EN to JP, the issue solved itself.
As far as I remember, I closed 3~10 tickets with IOException(not found) issue, at LEAST.
(due to my bad coding practice, no exception handling)

MathewSachin:
@29988122 I think the error must mean that you don't have `support_region_tool.png` in your `image_EN` folder.
Is your `image_EN` folder up-to-date? IIRC, that image was added recently.

29988122:
> 
> 
> @29988122 I think the error must mean that you don't have `support_region_tool.png` in your `image_EN` folder.
> Is your `image_EN` folder up-to-date? IIRC, that image was added recently.

......OK, I blame this software (for my stupid mistake)
圖片
* Only synced partially * (I sync github folder among emu and desktop by localhost ftp server)

Anyway, everything's solved, and thank you again for this tool!

potchy:
Isn it okay to merge this?

reconman:
I was waiting for the message change and then I forgot about it.

PR #368, opened 2020-04-22
Add images for TW server

aqer:
Current script can't work on TW server , so I add retry.png to fix it.
I also add withdraw.png and support_region_tool.png.


potchy:
thanks, very helpful :)

Issue #367, opened 2020-04-21
Assign Party to Script


kenwiel:
I just recently updated to the most recent version and noticed that its now possible to store multiple scripts.

As an extension of this, it would be useful to select which party is assigned to each script, so for XP farming use Party1, for QP Party 2, etc.

reconman:
Are you sure it's not already covered by the Autoskill_List? https://29988122.github.io/Fate-Grand-Order_Lua/#/features?id=autoskill

I'm running one script for multiple teams where I just have to select the support/skill configuration when I start the script.

MathewSachin:
I think they are talking about selecting the party in the game before the quest.

reconman:
Aaaah, that's going to be pretty difficult. OCR is definitely not supported, like detecting whether a team called "QP" is currently on the screen. You could do an image comparison of "PARTY II", "PARTY III", and so on.

The main problems with this feature are making it robust against the lags that come from party switching, dealing with user input validation, and also being able to run the feature on all server regions.

Now, FGO CN is currently not working at all with this script because they changed much stuff and we don't have any developer playing CN. And I haven't heard of any Taiwan player since I joined the project. So we'll probably only implement it correctly for JP and EN.

forgehe:
We could try clicking on the small party dot indicators on top instead of swiping, or clicking on the left/right buttons. But we may run into issues with different screen sizes and resolutions misclicking who knows.

PR #401, opened 2020-07-10
Feature/party selection

forgehe:
#367 

I'd appreciate it if ppl test that it works on their devices as well. 
The tapping locations are quite close to each other, so I don't want any misclicks to occur.
I'm mainly concerned about phones with notches, as I don't have a way to test that. 
Confirmed to work on Samsung S9.


forgehe:
Just ran into an issue, where if the selected party was already in use, the start quest click wouldn't register for some reason.
(Eg. AutoSkill uses Party 1, Party 1 is already the default. Start Quest Button doesn't register)
(Eg2. AutoSkill uses Party1, Party 2 is the default. Start Quest Button registers the click)

reconman:
I didn't even know that you can click on the dots at the top!

Seems like a good solution, I'll test this soon.

reconman:
@forgehe Can you check my corrections?

forgehe:
Corrections and documentation looks good!

reconman (approved):

forgehe:
Just had one occurance where it tapped to Party2, and 2nd tap to Party1 did not register. It may be a fluke, but I'm thinking it might occur more frequently on laggier devices.

